PR #3, opened 2012-03-13
Placeholder for WLineEdit

bytemaster:
Added ability to set the placeholder property.


emweb:
Patch is no longer relevant: this feature has been in Wt for a while now.

PR #23, opened 2013-04-10
Fix paths of background-image used in bootstrap theme.

pluggulp:
Hello Koen,

Please could you review this pull request and merge the changes into the upstream branch? The fix pertains to wrong paths for background-image under Bootstrap theme.

Thanks and regards,

~Plug


emweb:
Adding replaceWidget may still be a good idea, but this has been made a little bit more inconvenient in Wt 4. Maybe it could return a pair or a struct with two elements (which could work nicely with C++17's structured bindings).

In any case, this pull request is not applicable in this form anymore to the current code base so I'm closing it.

Regards,
Roel

PR #29, opened 2013-05-03
WMatchValidator support...

pluggulp:
Hello Koen,

Please review the code and accept this pull request for WMatchValidator support in Wt. A Match Validator matches two text fields. One of the use-cases for this validator is to compare two password fields during user account registration.

Kind regards,

~Plug


emweb:
While there may be some value to a WMatchValidator, I'm closing this PR since it is no longer compatible with Wt 4.

Regards,
Roel

PR #43, opened 2013-12-04
Fix unneeded dependency on boost.signals when boost.signals2 is selected

mologie:
Wt offers the option to use boost.signals2 as signals implementation, but always requests boost.signals as dependency, meaning that boost.signals must be built despite it not being used. This patch fixes said behavior by skipping the dependency if boost.signals2 is selected.

This required a little code shuffling, but appears to work well. I hope I did not miss anything - please make sure to test this with odd configurations before merging.


emweb:
This would have been helpful before the Wt 4 release, but since Wt 4 doesn't use Boost.Signals or Boost.Signals2 anymore, it's now obsolete. Sorry for not applying it way back when it was applicable. If you can solve the merge conflicts we can still get it into the wt3 branch if you really want it. I'm closing this PR for now.

Regards,
Roel

mologie:
Hi Roel, it's nice to hear that Wt's development is still going strong. I however don't plan on porting this change back to the wt3 branch. Thanks for your response.
Regards, Oliver

PR #52, opened 2014-03-07
Optional mutex on Transaction

tsaquet:
Replace previous pull request : https://github.com/kdeforche/wt/pull/46

Add the possibility to use Wt::Dbo::Transaction with a single Wt::Dbo::Session shared within multiple static resources. 

It is retrocompatible : the boolean added to manage the lock is optionnal and set to false by default.

Example : 
{
    // The transaction will be protected by the mutex
    Wt::Dbo::Transaction transaction(m_session, true);
    Wt::Dbo::Query<> queryRes = m_session.query<>("my_query");
    [...]
}


emweb:
What exactly is the use case for this feature? It seems to me that it's perfectly possible for the user of Wt::Dbo to add a lock whenever they would need it. What's the added value in integrating this into the Transactions?

Regards,
Roel

PR #60, opened 2014-06-29
Add Large File Support

pgquiles:
Add large file support (>= 4GB files with 32-bit binaries) on platforms that support it


emweb:
Hi, I'm not sure which issue this is trying to solve. Where would this be necessary exactly? Where does Wt not currently properly support larger files in some places when making a 32-bit build?

Regards,
Roel

PR #62, opened 2014-09-02
Running Wt.js through jshint

rgraebert:
This solves most of the issues raised by jshint:
- == instead of ===
- != instead of !==
- extra semicolons
- inner functions not at the beginning of outer functions
- duplicate variables
- constructors without ()

Remaining issues relate to eval.

### DIFF
--- a/src/web/skeleton/Wt.js
+++ b/src/web/skeleton/Wt.js
@@ -1,4 +1,4 @@
-
+/*jshint bitwise:false */
 /**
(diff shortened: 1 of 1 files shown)


emweb:
While we may want to update the Wt.js code to be in line with all JavaScript best practices (although opinions can vary wildly about what that means), this PR is now too old, and won't merge properly anymore.

Regards,
Roel

PR #65, opened 2015-01-22
WtDbo: Use SQLITE_TRANSIENT storage specifier for Sqlite blob binding

jtomson-mdsol:
When writing a specialization of ```sql_value_traits::bind``` where data is generated and stored in a local variable (i.e an encrypted field) I noticed that the persisted data in the db would often be corrupted.

In the sqlite backend implementation, ```SQLITE_STATIC``` was used as the storage specifier, which assumes that the bound data will remain allocated until the statement is finalized. However finalization occurs after the scope of the ```sql_value_traits::bind``` method.

I understand not wanting to create a temporary copy of blob data (which could be quite large) when binding, but this implicit requirement to handle allocation beyond the scope of ```sql_value_traits::bind``` is a nasty gotcha?

Thanks for taking a look!


kdeforche:
Hey, I believe the STATIC was unintended, actually!

PR #66, opened 2015-02-03
added mariadb to the lib path

pollend:
the lib path for mariadb was missing.

### DIFF
--- a/cmake/WtFindMysql.txt
+++ b/cmake/WtFindMysql.txt
@@ -28,6 +28,7 @@ FIND_LIBRARY(MYSQL_LIB
 	${MYSQL_LIBRARY}
     PATHS
     ${MYSQL_PREFIX}/lib
+    ${MYSQL_PREFIX}/lib/mariadb
     ${MYSQL_PREFIX}/lib/opt
     ${MYSQL_PREFIX}/lib/mysql
     ${MYSQL_PREFIX}/lib64/mysql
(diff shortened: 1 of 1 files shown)


kdeforche:
I agree, but would prefer to put the path in the end, to not break the setup of existing users. Would that still work for your case?

pollend:
i don't follow.

pollend:
i just tested it out, it still doesn't build correctly. I'll try mess around and see if i can get this to work else i might just use sqlite3.

pollend:
i ended up just downloading the connector from mariadb and working with that. -DFIMYSQL_PREFIX= is not really that clarifying as an error and the wiki doesn't give much help in terms of fixing this. might clarify the error with telling the individual that the connector library couldn't be found. or... you could include the connector library inside of the project as a sub-module.

JohnRobson:
pollend, the full parameter (in your case) is what? "-DFIMYSQL_PREFIX=?????"

pollend:
i downloaded the connector to my desktop.

JohnRobson:
Thank you, but, I skipped...  what's the connector name and sufix? "-DFIMYSQL_PREFIX=/path/conn.ini"

Could you post the page that you got thi

JohnRobson:
Thank you, but, I skipped...  what's the connector name and sufix? "-DFIMYSQL_PREFIX=/path/conn.ini"

Could you post the page that you got this connector? Tks.

pollend:
you can get it from there site. you have to untar the file, run cmake, make, and make install. this should install all the necessary libraries to local user directory.

https://downloads.mariadb.org/connector-c/

pollend:
i moved the path down to the bottom.

PR #67, opened 2015-02-04
Comment boost::optional

JohnRobson:
With these lines I got this error: http://paste.debian.net/144015/

I comment these lines and the compilation finish successful.

test/dbo/DboTest.C 
149:  //boost::optional<std::string> string3;
201:    //if (string3  != other.string3) {
202:    //  DEBUG(std::cerr << "ERROR: string3 = " << string3 << " | "
203:    //        << other.string3 << std::endl);
269:      //&& string3 == other.string3
296:    //dbo::field(a, string3, "string3");

I'm using GCC 4.9.2, Boost 1.57.0, cmake 3.1.1, Arch Linux 64 bits

### DIFF
--- a/test/dbo/DboTest.C
+++ b/test/dbo/DboTest.C
@@ -146,7 +146,7 @@ public:
   Wt::WString wstring2;
   std::string string;
   std::string string2;
-  boost::optional<std::string> string3;
+  //boost::optional<std::string> string3;
   boost::posix_time::ptime ptime;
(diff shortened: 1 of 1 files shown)


kdeforche:
Hey,

Unfortuantely the paste is gone; I'm too late so see it (that's shameful, yes). Do you still get the error and how does it look like?

JohnRobson:
Thank you Koen, since 3 months ago is working without problem.

PR #68, opened 2015-03-30
TIME_UTC problem workaround (boost 1.49, C++11)

starius:
Definition of TIME_UTC in boost/thread/xtime.hpp conflicts with glibc definition of TIME_UTC (C++11). Both of them define TIME_UTC = 1. This problem was fixed in boost 1.50, but Wt declares needed boost version >= 1.41, so it should compile against versions < 1.50 with modern g++ with C++11 support.

Example of error message:

```
In file included from /usr/include/boost/thread/pthread/mutex.hpp:14:0,
                 from /usr/include/boost/thread/mutex.hpp:16,
                 from /usr/include/boost/spirit/home/classic/core/non_terminal/impl/object_with_id.ipp:17,
                 from /usr/include/boost/spirit/home/classic/core/non_terminal/impl/grammar.ipp:15,
                 from /usr/include/boost/spirit/home/classic/core/non_terminal/grammar.hpp:21,
                 from /usr/include/boost/spirit/home/classic/core.hpp:42,
                 from /usr/include/boost/spirit/include/classic_core.hpp:11,
                 from /wt/src/http/StaticReply.C:8:
/usr/include/boost/thread/xtime.hpp:23:5: error: expected identifier before numeric constant
/usr/include/boost/thread/xtime.hpp:23:5: error: expected ‘}’ before numeric constant
```

gcc version 4.7.2

See https://bbs.archlinux.org/viewtopic.php?id=144593

### DIFF
--- a/examples/codeview/CodeSession.h
+++ b/examples/codeview/CodeSession.h
@@ -11,6 +11,7 @@
 #include <Wt/WSignal>
 
 #include <vector>
+#include "Wt/boost-xtime.hpp"
 #include <boost/thread.hpp>
 
 class CodeSession
--- a/examples/feature/broadcast/BroadCast.C
+++ b/examples/feature/broadcast/BroadCast.C
@@ -8,6 +8,7 @@
 #include <Wt/WMessageBox>
 #include <Wt/WServer>
 
+#include "Wt/boost-xtime.hpp"
 #include <boost/thread.hpp>
 
 /*
--- a/examples/feature/serverpush/ServerPush.C
+++ b/examples/feature/serverpush/ServerPush.C
@@ -9,6 +9,7 @@
 #include <Wt/WProgressBar>
 
 #include <iostream>
+#include "Wt/boost-xtime.hpp"
 #include <boost/thread.hpp>
 
 /*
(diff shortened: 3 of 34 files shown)


kdeforche:
Darn. I'm hoping there to be a less painful solution, perhaps by including <boost/thread/xtime.hpp> as very first file?

starius:
Unfortunately I don't know any less painful workaround for this problem. 

If file `boost/thread/xtime.hpp` from boost < 1.50 is included by a compiler with C++11 support, it would not compile because of `TIME_UTC` definition. The only solution I know is to prevent system `boost/thread/xtime.hpp` from being included by providing its copy (of course, with `TIME_UTC` problem fixed) and including it from all files including `boost/thread/xtime.hpp` indirectly.

I have rebased and updated the pull request. Following files required manual merging/updating:
  * test/testenvironment/TestEnvironmentTest.C
  * examples/wtwithqt/lib/DispatchThread.h

emweb:
Because this fix affects so much of Wt, and I don't envision anyone using a version of Boost that old with Wt 4, I'm closing this PR. We should look into what is really the minimum version of Boost we support these days.

Regards,
Roel

PR #70, opened 2015-05-25
Support for 'having' clause in queries

matiu2:
Allows Queries to have a 'having()' clause (similar to where() .. but for aggregate fields like 'count(x)' or 'sum(x)'.

### DIFF
--- a/src/Wt/Dbo/Query
+++ b/src/Wt/Dbo/Query
@@ -33,12 +33,13 @@ namespace Wt {
 	std::vector<FieldInfo> fields() const;
 	void fieldsForSelect(const SelectFieldList& list,
 			     std::vector<FieldInfo>& result) const;
-	std::pair<SqlStatement *, SqlStatement *>
-	statements(const std::string& where, const std::string& groupBy,
-		   const std::string& orderBy, int limit, int offset) const;
-	Session& session() const;
+        std::pair<SqlStatement *, SqlStatement *>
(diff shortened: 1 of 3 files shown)


kdeforche:
Hey matiu2, I like the addition, but I believe the implementation of 'addHavingBy()' is far too complex. Unlike groupBy(), there is no need for expanding a database object to all its fields as that would be illegal SQL syntax. That would simplify the implementation to look like:

    if (!having.empty())
        result += " having " + having;

matiu2:
Thanks Koen. Good insight there. I'll re-do it :)

matiu2:
Hi Koen,

I have redone it more simply as suggested. It also now supports using having() multiple times on the same query now. eg. myquery.having("sum(employee.id) > 4").having("sum(employee.id) < 6");

I force pushed and rebased to my repo, so just click on the above commit to see it: https://github.com/matiu2/wt/commit/5010ee98a9a7f96b439a259394797a358713354b

Thank for your time. All comments appreciated.

matiu2:
Hold on this one. I found a bug .. will redo again..

matiu2:
OK, it's good to go again. This line was the bug I just fixed:

https://github.com/matiu2/wt/commit/e9d18a61efb196d4d898a377bc60f8a67019df65#diff-84261d322b4759643b4c615756e52045R391

It occured when I tried to compile the blog example.

PR #71, opened 2015-07-08
Honour CMAKE_PREFIX_PATH, the cmake standard way to specify a prefix …

marco-m:
…path to find packages.

This works on any OS, but especially on MacOSX it allows the user to avoid picking up the stale system OpenSSL, by invoking: cmake -D CMAKE_PREFIX_PATH=/opt/local.

Similar results can be already achieved with USERLIB_PREFIX. The difference is that USERLIB_PREFIX is a wt-specific way, while CMAKE_PREFIX_PATH is the standard cmake way to do the same. Actually I wonder why USERLIB_PREFIX exists at all :-)

Forgot to mention that without this patch, the cmake configuration steps fails because cmake is confused: it has found both OpenSSL below /opt/local and the system one:

    cmake -DCMAKE_PREFIX_PATH=/opt/local ..

    [..]
    -- Configuring done
    CMake Warning at src/CMakeLists.txt:442 (ADD_LIBRARY):
      Cannot generate a safe linker search path for target wt because files in
      some directories may conflict with libraries in implicit directories:

        link library [libssl.dylib] in /usr/lib may be hidden by files in:
          /opt/local/lib

      Some of these libraries may not be found correctly.
    [..]

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,7 +234,13 @@ ELSE(WIN32)
     SET(CONFIGDIR "/etc/wt" CACHE STRING "Path for the configuration files")
   ENDIF( NOT DEFINED CONFIGDIR )
 
-  SET(USERLIB_PREFIX_DEFAULT "/usr")
+  # If the user specifies the standard CMAKE_PREFIX_PATH to find packages,
+  # honour it.
+  IF(CMAKE_PREFIX_PATH)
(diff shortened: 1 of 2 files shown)


marco-m:
Hi, any chance to have a look at this?

kdeforche:
Hey, yes, I will want to merge these but haven't had a chance to review the consequences in detail

PR #72, opened 2015-07-10
Fix cmake/FindWt.cmake to actually respect the REQUIRED flag.

marco-m:
Consider a `CMakeLists.txt` file used by a project that wants to use Wt:

    1 find_package(Wt REQUIRED)
    2 if(NOT Wt_FOUND)
    3     message(FATAL_ERROR "This should NOT happen!")
    4 endif()
    5 include_directories(${Wt_INCLUDE_DIR})

Since `find_package()` uses the `REQUIRED` flag, line 3 should never happen: the
code of `FindWt.cmake` should handle the `REQUIRED` itself and fail.

On the other hand, if you try to run the above snippet on a system either
without Wt or with Wt installed in a non-default location, you will see that
line 3 is actually executed.

Looking at FindWt.cmake, it is easy to understand why: all the logic is wrapped
by a `IF( Wt_INCLUDE_DIR )` guard, so if the Wt include dir is not found, it
will never handle the REQUIRED flag!

This patch does two things:

1. It removes the `IF( Wt_INCLUDE_DIR )` guard which is useless and confusing.
2. It fixes the indent in the whole file to be uniform (which was a mix of different spaces and sometimes tabs).


marco-m:
Hi, any chance to have a look at this?

PR #73, opened 2015-07-14
Fix FTBFS on arm64 due to 'char' meaning 'unsigned char' on arm64, wh…

pgquiles:
…ile 'signed char' on x86. Fixes Debian bug #791417


emweb:
This issue should no longer present in the latest master.

Regards,
Roel

PR #74, opened 2015-08-25
Use minified .css files

renindustries:
FontAwesome and Bootstrap's javascript is minified, and the files are all there; why are we wasting bandwidth when we don't have to?


kdeforche:
Hey, I agree related to the minified CSS. I can't merge this though since you also included other commits. What's TimePickerEx?

renindustries:
Woops! Let me issue a new pull, sorry about that

PR #80, opened 2015-10-13
Add option WT_FORCE_BOOST_DYN_LINK

png85:
This commit adds a build option `WT_FORCE_BOOST_DYN_LINK` which enforces `-DBOOST_ALL_DYN_LINK`
being added to compiler flags. While it's technically possible to abuse variables like `WT_CPP_MODE` for this or set `CXXFLAGS` environment variables this is a slightly cleaner approach.

The reason for this is that with some combinations of Boost / Wt build options a program using both Wt and Boost.Log will crash when log sinks & related objects get destroyed since the're are conflicting symbols from static and dynamic variants of `libbost_log_setup` being mixed in the same binary.


emweb:
Hi, do you know if this is this still an issue with Wt 4?

Regards,
Roel

PR #81, opened 2015-10-21
base64: fix invalid narrowing conversion

andreas-schwab:
Fixes this error:

src/web/base64.cpp:39:3: error: narrowing conversion of '-1' from 'int' to 'char' inside { } [-Wnarrowing]


emweb:
This issue should no longer be present in the latest master (made from_table signed).

Regards,
Roel

PR #83, opened 2015-12-17
Added missing functionality to FindWt.cmake:

SaiFi0102:
-WT_PREFIX cached path
-Definitions for MySQL and Firebird


emweb:
This has been obsoleted by Wt 4. I also believe that we already fixed most of this on the wt3 branch in the meantime.

Regards,
Roel

PR #91, opened 2016-04-08
WJavaScriptSlot: Pass slot's return value through to event caller

raytrap:
Being able to return values is at least useful to suppress key presses  by returning false in _onkeydown_. I have no idea about potential impact on other parts of Wt as I'm completely new to the library, but still this seems valid to me.


emweb:
I'm not sure if this is a good idea. You can connect multiple slots to one signal, so I think this will cause problems. In any case, you can use `preventDefault()` instead of returning false.

Regards,
Roel

PR #93, opened 2016-06-03
Added support for Forward Secrecy to wthttpd

terborg:
I've added support for forward secrecy to wthttpd. An option to set server cipher order preference is added. In addition, SSL_CTX_set_ecdh_auto, if available, is set to 1. Web servers such as Apache and nginx also set this to 1 by default.


terborg:
Deploying Wt through the wthttpd is not recommended. Rather, for production, use fcgi in conjunction with, e.g., nginx running ssl and http/2. This delivers the strongest ssl, no need for websockets, and faster connections through http/2.

PR #94, opened 2016-08-01
Update WMessageResources.C

bryongloden:
It's always good to initialize integer variables, at least to 0, because if you try to retrieve its value before it gets assigned any actual (non-garbage) value, then it results in undefined behavior, e.g. when the token WT_NO_SPIRIT is defined.

Found by https://github.com/bryongloden/cppcheck


emweb:
Thanks

PR #95, opened 2016-08-04
we must include API key as parameter in Google Maps API URI

k0bl:
A couple of things here. sensor=false is deprecated, I removed it to eliminate warning messages with the Version 3 API. 
As of June 2016, Google requires all maps API requests to be authenticated. 
The jsapi auth and google.load method we've been using previously results in a MissingKeyMapError. 
It appears we must include API key as the 'key' parameter when constructing the URI for the maps script tag. After doing so it loads fine with no error or warning messages.


emweb:
Closing this since it was fixed in commit c08ad28.

PR #96, opened 2016-08-16
fix cmake warning

juliendehos:
suppress the following cmake warning:

```
CMake Warning (dev) at test/CMakeLists.txt:42 (ADD_EXECUTABLE):
  Policy CMP0037 is not set: Target names should not be reserved and should
  match a validity pattern.  Run "cmake --help-policy CMP0037" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  The target name "test" is reserved or not valid for certain CMake features,
  such as generator expressions, and may result in undefined behavior.
This warning is for project developers.  Use -Wno-dev to suppress it.
```


wdu:
This is fixed by changing the target name to test.wt

PR #98, opened 2016-09-30
Add autotools to compile hello example and composer

juanfe:


wdu:
Juan,

I'm a bit lost at how this improves Wt's build system. Wouldn't it be more appropriate to add a 'How to build your Wt application using autotools' in the Wt Wiki?

Wim.

PR #99, opened 2016-10-27
Add AlwaysShowEditors EditOption

shiftee:
This causes a WTableView to render the editor for any editable indexes.
It also adds the editor to the internal editedItems_ container which enables editing of the model.

This pull request is about the following comment in this thread:
[http://redmine.webtoolkit.eu/boards/2/topics/9148](http://redmine.webtoolkit.eu/boards/2/topics/9148)

> If you always want to show the combobox, specialize WAbstractItemDelegate, and implement update() so that it always renders a combobox.

I am unable to get a custom delegate based directly on WAbstractItemDelegate to call setModelData().
After looking at the code I think this is because WAbstractItemView::closeEditorWidget()
does not find the editor in its editedItems_ container. This is because editing has not been triggered.
This request therefore triggers editing automatically by calling the edit() function.

Have I missed something??

### DIFF
--- a/src/Wt/WAbstractItemView
+++ b/src/Wt/WAbstractItemView
@@ -57,10 +57,11 @@ public:
    * \sa setEditOptions()
    */
   enum EditOption {
-    SingleEditor = 0x1,    //!< Never show more than one active editor
-    MultipleEditors = 0x2, //!< Allow multiple editors at the same time
-    SaveWhenClosed = 0x4,  //!< Always save the current edit value when closing
-    LeaveEditorsOpen = 0x8 //!< Editors can only be closed using closeEditor()
+    SingleEditor = 0x1,      //!< Never show more than one active editor
+    MultipleEditors = 0x2,   //!< Allow multiple editors at the same time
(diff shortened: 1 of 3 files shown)


emweb:
I don't know which is more appropriate: this option or an option which makes editors initially open. This option seems a lot simpler to implement, though.

Regards,
Roel

shiftee:
Hi Roel,
Initially-open could suit where the user must choose each item once (e.g. flight search form).
I was thinking of a digital outputs menu where e.g. lots of toggle-switches are visible or sliders for analog outputs.
I guess it depends on the end-user workflow.

My priorities were changed soon after sending the pull request and I am not sure if it is working properly, I will test it further on Monday

shiftee:
I've just tested it again, it seems to be working fine.
However I am not overly familiar with the Wt code so there may be trivial errors.
Could you check that the quoted text and paragraph below it in the first post above are correct?

Thanks,
Mark

emweb:
Closing this due to inactivity.

PR #100, opened 2016-11-16
Add title to WGooglemap icon markers

raphaelbergamini:
Remove sensor option from WGooglemap


src/Wt/WGoogleMap
```diff
@@ -152,7 +152,7 @@ public:
 
   /*! \brief Adds a icon marker overlay to the map.
    */
-  void addIconMarker(const Coordinate &pos, const std::string& iconURL);
+  void addIconMarker(const Coordinate &pos, const std::string& iconURL, std::string toolTip = NULL);
```
wdu: Since tooltip is text that should be translatable, the following signature is more appropriate:
void addIconMarker(const Coordinate &pos, const std::string& iconURL, const Wt::WString toolTip = Wt::WString());

PR #101, opened 2016-12-07
Fix WTimePicker segfault on construction

shiftee:
Constructors which do not take a WTimeEdit as parameter segfault
due to dereferencing uninitialised timeEdit_ variable.
Issue fixed by default initialising timeEdit_ in these cases

### DIFF
--- a/src/Wt/WTimePicker.C
+++ b/src/Wt/WTimePicker.C
@@ -18,6 +18,7 @@ LOGGER("WTimePicker");
 WTimePicker::WTimePicker(WContainerWidget *parent)
     : WCompositeWidget(parent),
       selectionChanged_(this),
+      timeEdit_(new WTimeEdit()),
       toggleAmPm_(2, this)
 {
     init();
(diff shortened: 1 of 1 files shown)


shiftee:
Is this a valid use case?
If so should I redo it to delete the WTimeEdit only in those two cases, or should WTimePicker take ownership and always delete it?
The documentation for the WTimePicker constructors could also be updated to specify ownership.

emweb:
The WTimePicker wasn't really intended to be used without a WTimeEdit. Documenting it was probably a mistake. Or do you really have a use case where you would like to use a WTimePicker without a WTimeEdit?

Regards,
Roel

shiftee:
I am implementing a touchscreen system's settings menu.
I am using a WTableView but each row may have different types (time,float,string etc.).
The delegate renders each row as a string (WItemDelegate) but createEditor() opens a dialog
containing an editor of the correct type, along with Cancel & Done buttons.

In this case it is preferable to have a WTimePicker already open instead of a WTimeEdit where the user
needs to click the edit button to begin editing.

WTimePicker only uses the WTimeEdit for formatting purposes.
In my opinion it would be much cleaner and more modular to remove WTimeEdit from WTimePicker and add simpler api calls for setting these options. WTimeEdit could then be refactored to use these.
It would also fix the confusing current situation where WTimePicker contains a WTimeEdit member and WTimeEdit contains a WTimePicker member
dialog_time_edit

shiftee:
Or perhaps constructors which do not take a WTimeEdit as parameter should be removed?

emweb:
We've concluded that we'll simply remove the documentation from WTimePicker, since it was only intended to be part of the implementation of WTimeEdit.

PR #88, opened 2016-01-22
Patch: support HTTP PATCH in http/Client and http/RequestHandler

jamal-fuma:

PR #104, opened 2017-02-10
HTTP PATCH support for Wt 3.7

jamal-fuma:
This patchset adds support for making Http Patch requests using Wt::Http::Client.
This patchset allows WResource to respond to Http Patch requests.
This patchset expands Wt::Http::Redirection support to allow for Method preserving redirection - e.g. Http 307/308 with a POST results in a Post.
This patchset expands Wt::Http::Redirection support to allow for Method translating redirection - e.g. Http 301/302 with a POST results in a Get.

### DIFF
--- a/src/Wt/Http/Client
+++ b/src/Wt/Http/Client
@@ -285,7 +285,22 @@ public:
    * \sa request(), done()
    */
   bool deleteRequest(const std::string& url, const Message& message);
-  
+
+  /*! \brief Starts a PATCH request.
+   *
+   * The function starts an asynchronous PATCH request, and returns
+   * immediately.
+   *
+   * The function returns \c true when the PATCH request has been
+   * scheduled, and thus done() will be emitted eventually.
+   *
+   * The function returns \p false if the client could not schedule
+   * the request, for example if the \p url is invalid or if the %URL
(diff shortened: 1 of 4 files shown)


emweb:
When automatic redirection was implemented, a conscious decision was made only to automatically redirect GET requests, and leave the decision to follow redirects on POST, PUT and DELETE to the user of the API. I'm not sure if we should do this automatically for POST, PUT and DELETE requests.

jamal-fuma:
The redirect for non GET requests should really have been in a separate pull request, I wanted this one to be PATCH support.

Nevertheless  the RFC does indeed say a user agent MUST NOT redirect unless confirmed. However the current client doesn't warn on 307/308 redirects, and doesn't follow the redirect, which breaks some of my uses of Wt::Http::Client.

Perhaps an additional method - followRedirectsWithoutConfimation(bool) would be sufficient

jamal-fuma:
The original pull request without redirection support is https://github.com/emweb/wt/pull/88

emweb:
PR #88 wouldn't properly merge. I've done it manually instead and will push the changes soon.

jamal-fuma:
Brilliant thanks very much

PR #107, opened 2017-03-31
Allow developer to tell cmake where Wt might have been installed via use of Wt_ROOT variable.

keith-bennett-gbg:
Allow developer to tell cmake where Wt might have been installed via use of Wt_ROOT variable.

Example use:
```
git clone git@github.com:em/wt.git && cd wt && mkdir build && cd build && cmake -DCMAKE_INSTALL_PREFIX="${HOME}" ../ && make -j$(nproc) && make install
cd ~/my_project/ && mkdir build && cd build && cmake -DWt_ROOT="${HOME}" ../ && make -j$(nproc) && make install
```

Alternative use:
```
export Wt_ROOT="${HOME}"
git clone git@github.com:em/wt.git && cd wt && mkdir build && cd build && cmake -DCMAKE_INSTALL_PREFIX="${Wt_ROOT}" ../ && make -j$(nproc) && make install
cd ~/my_project && mkdir build && cd build && cmake ../ && make -j$(nproc) && make install
```


wdu:
Good feature. Shouldn't the PATHS ${Wt_Root} PATH be changed to simply PATHS ${Wt_Root}? I can't imagine what that extra 'PATH' is doing there.

keith-bennett-gbg:
Frankly CMake isn't exactly my favored language.

The way I'm parsing this when I read it and [consulting the documentation](https://cmake.org/cmake/help/v3.0/command/find_library.html) is:
`PATHS` tells `FIND_LIBRARY` where to search. If there's more than one path, then it will look in the same order provided. My understanding is that `PATH` then tells it to look in the variable named `PATH`, which is the environment `PATH` variable (to find system-installed libraries). So it ends up using `Wt_ROOT` first. If that's empty or doesn't have the expected libraries, _then_ it would search `PATH` variable.

If it doesn't belong there, then why was it there in the first place? I don't know enough about CMake to answer that. Please feel free to enlighten me :)

emweb:
I think it should've said "ENV PATH", but it was mistakenly set to just "PATH". I'm not sure if "ENV PATH" really has a use there, as include or library paths are usually not in the PATH environment variable anyway, so I believe it can be simply left out.

Regards,
Roel

emweb:
This has been obsoleted by Wt 4's move to using CMake's export functionality, i.e. find_package using Config mode rather than Module mode.

PR #109, opened 2017-04-12
Hunterize Wt

caseymcc:
This is not complete yet, more to get a discussion started if this is something you would allow in the Wt project.

This adds support for hunter (a cmake package handler, https://github.com/ruslo/hunter). Hunter will fetch the required packages and compile them as needed. Currently for this it will get the following

boost
MySQL
libpng
OpenSSL
zLib

I don't think packages exist for the remaining but can be added. If you would like you can read the documentation for hunter but the main gist is that hunter will allow the end developer just to run cmake and the packages will be fetched (and compile) based on the toolchain that is in use. It keeps track of the compiles by toolchain (and options) so in general it allows for easy cross compiling without the head ache of local libraries butting in.

I plan to set Wt up as a package itself and if I can get these items into the main trunk then I can keep from having to maintain a separate repo for the hunter version. I started with wt4 but should be relatively easy to put this in 3.X.

Let me know your thoughts and any suggestions.


notorand-it:
I don't think that will help much.

0. I used to prefer pre-compiled packages for wt until I hit a few bugs that have been fixed in newer releases that are lagging behind in my distro of choice (Ubuntu server 16.04.2). Now I just recompile wt, not all of its dependencies.

1. Packages do actually exist for all the needed dependencies. At least for Ubuntu 16.04.2 and, I would bet, also for all major distributions. I have no idea about Windows, though.

2. The configuration step clearly highlights the unmet dependencies. In case those are blocking ones, the admin that's in charge will very likely prefer the stock packages from the distro over the locally compiled ones.

In any case, if the "great chiefs" will decide to go with hunter, I would like to have it as an option, not a mandatory dependency for compilation: I am not sure on how many distros already package hunter in their repositories.

wdu:
Hey casey,

Among the many options for package management available, it's hard for me to commit on supporting hunter (or any other one that's not somewhat a de facto standard) inside Wt's build system. As a matter of fact, this is the first package system I encounter that would require a project to integrate it in its build system. Is this really a requirement? Wt's configuation should be flexible enough to accomodate package manager's requirements... Can't the package manager use something that wraps around Wt's build system (note that I don't know hunter at all)? How is e.g. boost packaged in hunter, which is a project that doesn't even use cmake?

I'll be more willing to add more flexibility in the WtFind... scripts so that dependencies can properly be found, if the current implementation is insufficient for one or another packaging system.

Wim.

caseymcc:
I tried to minimize the impact of adding hunter and left it as an option that defaulted to off, but I fully understand that supporting ever package manager (especially ones that require modifications to ones build configurations) would be daunting.

By know means is hunter the current de facto package manager, it think it is relatively new. I can't give you detailed reasoning on the ins and outs of it as I am also new to it. And there are things about it that I wished work a little different. However I can give the reason I use it and why I have made efforts to move anything I use into it. First is is fully cmake based, I dont have any other software/interpreter/scripting/etc... I need to run it. I already use cmake and if cmake is not considered the de facto yet I think it soon will be (I mean Micrsoft built cmake support into VS17 and we know how they are) as many things are slowly moving to it. Next I work mainly on Windows (my addiction to MSVC's debugger is hard to break). As you know there is not an OS level package handler there no apt-get, yum, etc... so these package handler do me little use. Third, hunter allows the project to call out version/options of the libraries as well as including the toolchain as an identifier. It sperates the builds based on toolchains/versions/options and keeps them in separate locations allowing for easier cross compiling. The projects using hunter are directed at cmake time to the correct folder depending on toolchain in use and the version/options of the package. This keeps me from wring the normal documentation on how to compile.

    in order to use this project you need the following
    libraryA
    libraryB , make sure to compile with optionA off
    libraryC, only use version 3,X, 2.X is not supported
    etc...
    cmake -H.- B_build
    cmake --build _build

instead documentation just says 
    run 
    cmake -H.- B_build 
    cmake --build _build

and generally reduces the "I can't compile issues" from end users


On to your questions as best that I can answer:

"this is the first package system I encounter that would require a project to integrate it in its build system. Is this really a requirement?"
    Yes, as mention this all runs inside of your cmake build, HunterGate needs to be included to know how to fetch the Hunter (including which release). Once hunter is download from its github repo the call to hunter_add_package handle all the leg work of fetching the version of the library requested, setting up the  options and building it, either through cmake or whatever build system the library uses.

"Can't the package manager use something that wraps around Wt's build system (note that I don't know hunter at all)?"
   Yes and no, as mentioned it needs to be part of Wt's cmake. However they way projects are handled that do not include hunter in their build setup, a separate repo is setup and a clone of the original is made with the required modifications under a hunter branch (which can be seen here https://github.com/hunter-packages). The same thing is currently being for Wt https://github.com/hunter-packages/wt. I have just engaged you as including hunter directly makes things easier on the back-end and make the Wt compile easier for end users. And if ever hunter is ever to be the de facto it needs to be put places :)

"How is e.g. boost packaged in hunter, which is a project that doesn't even use cmake?"
    Hunter includes cmake scripts specifically for boost in order to call its build system. Hunter generally can download source from a released zip (whether that be github or some know location) then build with cmake or custom script if required. These are the packages it supports so far (and more seem to be added daily) https://docs.hunter.sh/en/latest/packages/all.html.

I am currently testing the modification for Wt3.X and will then move them into the hunter-packages repo to allow Wt to be setup as a package. Once that is complete I can let you know and you can play with it to see if you like it. You can test the Wt4 version now but I have only tested it on Windows, once it is in the hunter-packages its travis/appveyor testing will kick in and I will know if there are issues on other platforms.

s-martin:
I didn't grasp yet Hunter's advantage over CMake's built in Add_ExternalPackage.

PR #111, opened 2017-04-13
Add support for forward secrecy to wthttp

terborg:
I've added support for forward secrecy to wthttpd. An option to set server cipher order preference is added. In addition, SSL_CTX_set_ecdh_auto, if available, is set to 1. Web servers such as Apache and nginx also set this to 1 by default.


wdu:
Thank you!

PR #113, opened 2017-04-28
ability to add a logo image to navigation bar

bold84:


ordisbold:
Oh, I forgot to mention the advantage over just adding the image as a widget; the logo will be visible even when the menu is collapsed (on mobile devices), which is not the case with widgets that have been added via WNavigationBar::addWidget().

PR #114, opened 2017-05-03
Workaround issue when compiling with Boost 1.64

keith-bennett-gbg:
Workaround issue when compiling with Boost 1.64. See here: http://redmine.webtoolkit.eu/boards/2/topics/13412


emweb:
Thanks!

Roel

PR #118, opened 2017-07-10
Make WSortFilterProxyModel handle src-model-reset

shiftee:
The filter-proxy-model now emits its modelReset signal when its
source model does so

http://redmine.webtoolkit.eu/boards/1/topics/12567

Is this behaviour correct?

I am currently unable to reproduce the issue which required a call
to invalidate() or resetMappings(), perhaps this should be checked
if you decide to merge


emweb:
That doesn't seem to be the only proxy model that doesn't properly handle modelReset(). I implemented it for all of the proxy models where it was missing and will push that fix later. Your implementation is not sufficient, because you also need to reset the mappings.

Regards,
Roel

PR #119, opened 2017-08-03
Add missing header

krzysztofwos:


emweb:
It seems that the include is indeed technically necessary, even though it was compiling for us without it. On what distribution did it fail to compile? I fixed it in commit b2d97de.

krzysztofwos:
This happened on macOS. I guess another header on Ubuntu or other Linux distros must have included `select()`. Always include what you use :)

emweb:
Of course :-).

Weird, for me it compiled on macOS without that include.

Regards,
Roel

krzysztofwos:
Not that it really matters, but here's compiler/system info:
```
$ clang --version
Apple LLVM version 8.1.0 (clang-802.0.42)
Target: x86_64-apple-darwin16.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
$ uname -a
Darwin MacBook-Pro-2016.local 16.6.0 Darwin Kernel Version 16.6.0: Fri Apr 14 16:21:16 PDT 2017; root:xnu-3789.60.24~6/RELEASE_X86_64 x86_64
```

PR #122, opened 2017-10-30
Added support for `std::chrono::seconds`

arteniioleg:
Fixes https://redmine.webtoolkit.eu/issues/6075


arteniioleg:
I wouldn't bother you if there will be a possibility to add this functionality without changing the core files.

Can you merge this please?

emweb:
This PR seems incomplete: you only implement getResult, and only for MySQL, PostgreSQL and SQLite, not for SQL Server and Firebird.

You could actually implement a trait that simple saves the count() as a large integer, without needing to change Wt.

I agree that it's inconvenient that we don't support std::chrono::duration outside of std::chrono::duration<int, std::milli> at the moment. I'll see what can be done about that. One nuisance is that while many databases support time types, many of those time types are rather limited (e.g. only to a single day, or a few days), so a large integer would then maybe be the only globally appropriate type.

Regards,
Roel

arteniioleg:
Thanks

PR #124, opened 2018-02-06
Compile-fix for when CSS_PARSER is not defined

frodegill:
One of my machines (2GB RAM, no swap) runs out of memory compiling CssParser.C. For Wt3 I could undefine CSS_PARSER in that file and compile a working library. For Wt4, I get a compile error (seems like someone forgot to upgrade this part of the code). Provided is a compile-fix.

### DIFF
--- a/src/Wt/Render/CssParser.C
+++ b/src/Wt/Render/CssParser.C
@@ -501,22 +501,22 @@ namespace Wt {
 CssParser::CssParser()
 { }
 
-StyleSheet* CssParser::parse(const WString& styleSheetContents)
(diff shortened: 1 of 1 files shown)


emweb:
We indeed did not test if it compiled with CSS_PARSER not defined. It's a hidden feature anyway, one of which I wasn't if sure we need it anymore. It's indeed a file that requires a lot of memory to compile, though. I didn't think people were still trying to compile C++ code on machines with 2 GiB RAM and no swap.

I'll take a look at this PR, but I don't know if our intention is to really keep supporting that unless it's really necessary to get it to compile for certain platforms.

Regards,
Roel

PR #126, opened 2018-04-04
Corrected documentation about ownership transfer

limitedAtonement:
With Wt 4, ownership is made explicit, so no comment need be made
about how ownership works, or under what circumstances a model
will be deleted.


emweb:
That was indeed left over. Thanks.

Roel

PR #127, opened 2018-04-04
Correcting typo and remove example comments

limitedAtonement:
This code doesn't look like C at all.


emweb:
Yeah, Koen uses Emacs, so those are actually not really directed to users, but to Emacs, so it interprets the file as C++.

Regards,
Roel

PR #130, opened 2018-04-27
Documentation was incorrect for C++

limitedAtonement:
This appears to act differently in C++ vs Java as WT_USTRING is different
in the two environments. The documentation now reflects that.


emweb:
I fixed it in a slightly different way.

Regards,
Roel

PR #132, opened 2018-06-06
Fix 500 status leading to invalid HTML reply [master]

GoupilRobin:
A missing break was breaking the HTML reply by inserting two status lines instead of only one


GoupilRobin:
FYI, here is the faulty commit: 0fe2d3a67d0d414ae26a6477e102b3e8c8ed01ff, introduced in 3.3.5

PR #134, opened 2018-06-17
Rate limit scrolling history.replaceState

hyperair:
Scrolling on Chrome Beta for Linux with native smooth scrolling generates a pretty high frequency of scroll events which results in choppy scrolling behaviour. Chrome Inspector shows a repeated message of:

```
Throttling history state changes to prevent the browser from hanging.
```

This PR rate-limits the `updateScrollHistory` call to avoid those stalls.


wdu:
Thanks for the patch.

Would it be possible to modify the patch so that the arguments of the last invocation of proxy() before the timeout are used instead of the arguments at the time of creation of the timer? I don't trust my JS skills enough to simply move the 'var args = arguments;' line to above the 'if (timer)' test, but do you get what I mean?

Wim.

hyperair:
Alright, done.

emweb:
I minified Wt.min.js myself and did a manual merge of this patch to both the wt3 and master branches.

Thanks,
Roel

hyperair:
On Tue, Jun 19, 2018 at 03:42:25AM -0700, emweb wrote:
> I minified Wt.min.js myself and did a manual merge of this patch to
> both the wt3 and master branches.

Great, thanks!

-- 
Kind regards,
Loong Jin

PR #135, opened 2018-06-26
Fix WStringStream to accept const char strings

zachstruck-ordis:
Fix `WStringStream` to directly accept `const char *` strings.  
Remove overly greedy template overload `operator<<`.  
Remove string literal overload to reduce template bloat since there is now a `const char *` overload that can accept string literals.


emweb:
Now to comment on the correct PR:

That's indeed a bit of a silly optimization. I noticed that the compiler will just optimize the strlen call out if the implementation is in the header. I updated your PR with that little change.

Regards,
Roel

PR #137, opened 2018-06-27
Fix 6450

apivan2:
For the first loaded WMenuItem, loadContents() was executed before connectSignals() and moved uContents_ out. As a result, a call to implementStateless() was missing.
After this change, the tab will now able to be switched after the application is killed.

Without a call to implementStateless(), switching the tab require full client-server round-trip.

### DIFF
--- a/src/Wt/WMenuItem.C
+++ b/src/Wt/WMenuItem.C
@@ -388,7 +388,7 @@ void WMenuItem::selectNotLoaded()
 
 bool WMenuItem::contentsLoaded() const
 {
-  return uContents_ != nullptr;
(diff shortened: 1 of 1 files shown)


Apivan:
https://redmine.webtoolkit.eu/issues/6450

emweb:
Hm, I think there's more going on there. I would expect contentsLoaded() to be true if there is no uContents_, but there is an oContents_ (so when the contents are actually in the widget tree instead of owned by the WMenuItem). I'm looking into several related WMenu and WMenuItem bugs at the moment. This PR did help me to spot that oddity so thanks.

Regards,
Roel

PR #138, opened 2018-08-03
Basic PLAIN authentication for mail Client

sikora84:
Very basic PLAIN authentication mechanism for mail Client. It was needed on server which I was working on. Can be extended for other auth mechanism in the future. For this to work smtp-plain-auth property with base64 hash have to be set in wt_config.


emweb:
Closing this since authentication was implemented in Wt 4.3.0.

PR #141, opened 2018-09-17
Load javascript <script> with type="module"

limitedAtonement:
I'm guessing this isn't acceptable as is, but it may be a starting place if you want to think about supporting users' javascript with ES6 modules.


emweb:
I think we'll have to take a closer look at modules ourselves first, to know if this would be the right way of going about it.

I think it's not a good idea to add another string argument between two existing string arguments. It's a pretty subtle API-breaking change, and it makes an already pretty stringly-typed method even stringier. Maybe a type as an enum class would work, or adding flags, or we'll have to start looking at combining all of these arguments into an object. Another thing I've thought about is maybe adding support for the integrity attribute: https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

Regards,
Roel

limitedAtonement:
I was hesitant about adding a parameter to the middle. I went with it because after reviewing the code, the last parameter is never sent (if I remember correctly!), so it was much simpler to add this module parameter to the end of several call sites rather than adding an `undefined` then the module parameter.

Thanks for checking up!

emweb:
It looks like this `symbol` argument indeed isn't used. That's odd. I'll have to check that.

emweb:
Ah, it is indeed being used. `symbol` is a JavaScript expression. If it is defined, then Wt won't load that JavaScript library, since it will assume that it's already been loaded.

Regards,
Roel

limitedAtonement:
Feel free to close this if it isn't on the roadmap :-) I'm using it successfully to run some ES6 javascript at https://brightlight.today/speedlines/.

emweb:
I think we're not going to merge this as-is, but it is something to keep in mind. [I see there's also a Redmine issue that someone made](https://redmine.webtoolkit.eu/issues/6623).

Regards,
Roel

PR #142, opened 2018-09-24
Fix mysql backend build

akapust1n:
This fix helps to specify full path to the mysqlclient library.

For example, on ubuntu 18.04 the path to mysqlclient  is `/usr/lib/x86_64-linux-gnu/`.
This means that in release 4.0,4 I can not build mysq backend. My request helps to fix this issue.

### DIFF
--- a/cmake/WtFindMysql.txt
+++ b/cmake/WtFindMysql.txt
@@ -27,6 +27,7 @@ FIND_LIBRARY(MYSQL_LIB
     NAMES
 	${MYSQL_LIBRARY}
     PATHS
+    ${MYSQL_PREFIX}
     ${MYSQL_PREFIX}/lib
     ${MYSQL_PREFIX}/lib/opt
     ${MYSQL_PREFIX}/lib/mysql


emweb:
So are you disabling the default find locations? I can add it, but I'm not sure why it doesn't work for you.

Regards,
Roel

akapust1n:
I do not disable default find locations.  They will work as before. 

If mysqlclient library has not-standard path, then we can't to build wt mysql backend. 
My request added possibility to specify full path to the mysqlclient library for build.

emweb:
I just don't really see in which scenario this is applicable. `/usr/lib/x86_64-linux-gnu/` is a default library search path on Ubuntu, so it doesn't need to be listed.

If you add `${MYSQL_PREFIX}` to the `PATHS` when looking for the library, then you'd have the situation where it looks for the library's headers in an `include` directory inside of the lib folder?

If the include dir and lib dir are in an unusual location, then I think it may be better to just use `MYSQL_LIB` and `MYSQL_INCLUDE` to specify the location of the library and headers, or using `CMAKE_LIBRARY_PATH`.

Your PR suggests that you actually had some problem with this, though, so I do wonder what that was, then? Is there some scenario we're really missing?

I would prefer to a use standard CMake find module for this, but there doesn't seem to be one.

emweb:
Closing this due to inactivity.

PR #144, opened 2018-10-11
fix:usr: WPopupFilter was breaking if it got reparented because it wo…

matiu2:
…uld loose all jQuery data attached to the node

Now it'll store the data in a temporary variable before reparenting it
and restore it after it's reparented.

This would also affect WDateEdit and other widgets that store data on
dom nodes.


emweb:
I think this is no longer an issue, since we're not using jQuery.data anymore.

PR #145, opened 2018-10-30
struct WT_API Wt::NoClass

quatmax:
On windows with dll usage of WT you have to dllexport/dllimport Wt::NoClass.
Otherwise you get linkage erros...

### DIFF
--- a/src/Wt/WSignal.h
+++ b/src/Wt/WSignal.h
@@ -23,7 +23,7 @@ class SlotLearnerInterface;
 class WStatelessSlot;
 class JavaScriptEvent;
 
-struct NoClass
+struct WT_API NoClass
 {
   NoClass() { }
   NoClass(const JavaScriptEvent&) { }


emweb:
Is this warning C4251? When the compiler and flags match, the WT_API is not strictly necessary there. The compiler also warns about STL types like std::string etc. I can add WT_API to the types we define ourselves, but that doesn't eliminate the issue for the STL. Or are you actually really getting errors instead of warnings?

quatmax:
This was a warning while linking against wt.dll on windows.
It's not really a problem so I will close this pr.
I'm sure there are more important issues 🙂

quatmax:
Hi! 
I just had the problem again, linking against wt 4.0.5 (current vcpkg port-version).
Windows 10 64bit
Visual Studio 2019 16.3.4

`error LNK2001: unresolved external symbol "public: static struct Wt::NoClass Wt::NoClass::none" (?none@NoClass@Wt@@2U12@A)`

So maybe this is interesting for you...

emweb:
I see, it's the static instance `Wt::NoClass::none` that's giving you linker errors, if you e.g. try to `emit()` an `EventSignal<>`. Seems to me that it would be best to just get rid of `none` altogether and just use a default-constructed `NoClass`, or employ some other method of allowing 0 or 1 template arguments.

Regards,
Roel

emweb:
I pushed a commit to master that should fix this (so it does not actually need `NoClass::none`).

Regards,
Roel

quatmax:
Hi Roel!

Perfect, thanks a lot.

BR, Max

PR #146, opened 2018-11-01
Updating non-default WLineEdit constructor to match default constructor

JackMaguire:
The two WLineEdit constructors result in different behavior which isn't obvious to the user. The default constructor sets `setInline(true)` and `setFormObject(true)` but these settings are not applied when the user passes a string. This PR adds those two lines to the second constructor.


emweb:
This has already been fixed in the meantime.

PR #151, opened 2018-12-21
Wt3 - WTextEdit return just a html formatted String!

CMLDMR:
WTextEdit Widget return only html formatted text. there is no member to return plainText.


emweb:
This PR has gone a bit weird it seems? It wants to merge all of wt4 into wt3 or something?

Regards,
Roel

emweb:
I'm just going to close this due to inactivity.

Regards,
Roel

PR #152, opened 2019-02-06
Fix issue #6884

Apivan:


emweb:
This is fixed by commit 98b8986.

PR #154, opened 2019-05-09
Improved performance of adding rows and columns to the WTable

jakub-w:
The issue was the calling of `WTableRow::rowNum()` for every added cell,
which iterated over the whole table.

The methods `WTableRow::createCell()`, `WTableRow::insertColumn()` and
`WTableRow::expand()` got the new row argument defaulting to -1 that
`WTable` can use when creating new elements with `WTable::insertRow()` and
`WTable::insertColumn()`. These methods are called by `WTable::expand()`
so the performance improvement affects every operation that expands
the table.

The performance of populating a new table row by row is increased
with this commit from O(n^2+n) to O(n).

### DIFF
--- a/src/Wt/WTable.C
+++ b/src/Wt/WTable.C
@@ -103,7 +103,7 @@ WTableRow* WTable::insertRow(int row, std::unique_ptr<WTableRow> tableRow)
     widgetAdded(cell.get());
   }
   rows_.insert(rows_.begin() + row, std::move(tableRow));
-  rows_[row].get()->expand(columnCount());
+  rows_[row].get()->expand(columnCount(), row);
   repaint(RepaintFlag::SizeAffected);
 
   return rows_[row].get();
@@ -113,7 +113,7 @@ WTableColumn* WTable::insertColumn(int column,
 				   std::unique_ptr<WTableColumn> tableColumn)
 {
   for (unsigned i = 0; i < rows_.size(); ++i)
-    rows_[i]->insertColumn(column);
+    rows_[i]->insertColumn(column, i);
 
   if ((unsigned)column <= columns_.size()) {
     if (!tableColumn){
--- a/src/Wt/WTableRow.C
+++ b/src/Wt/WTableRow.C
@@ -26,19 +26,23 @@ WTableRow::WTableRow()
 WTableRow::~WTableRow()
 { }
 
-std::unique_ptr<WTableCell> WTableRow::createCell(int column)
+std::unique_ptr<WTableCell> WTableRow::createCell(int column, int row)
 {
-  if (table_)
-    return table_->createCell(rowNum(), column);
+  if (table_) {
+    if (-1 == row) {
+      row = rowNum();
+    }
+    return table_->createCell(row, column);
+  }
   else
     return std::unique_ptr<WTableCell>(new WTableCell());
 }
 
-void WTableRow::expand(int numCells)
+void WTableRow::expand(int numCells, int row)
 {
   int cursize = cells_.size();
   for (int col = cursize; col < numCells; ++col) {
-    cells_.push_back(createCell(col));
+    cells_.push_back(createCell(col, row));
     WTableCell *cell = cells_.back().get();
     if (table_)
       table_->widgetAdded(cell);
@@ -47,9 +51,9 @@ void WTableRow::expand(int numCells)
   }
 }
 
-void WTableRow::insertColumn(int column)
+void WTableRow::insertColumn(int column, int row)
 {
-  cells_.insert(cells_.begin() + column, createCell(column));
+  cells_.insert(cells_.begin() + column, createCell(column, row));
   WTableCell *cell = cells_[column].get();
   if (table_)
     table_->widgetAdded(cell);
--- a/src/Wt/WTableRow.h
+++ b/src/Wt/WTableRow.h
@@ -136,10 +136,10 @@ public:
   virtual const std::string id() const override;
 
 protected:
-  virtual std::unique_ptr<WTableCell> createCell(int column);
+  virtual std::unique_ptr<WTableCell> createCell(int column, int row = -1);
 
 private:
-  void expand(int numCells);
+  void expand(int numCells, int row = -1);
 
   WTable *table_;
   std::vector<std::unique_ptr<WTableCell> > cells_;
@@ -151,7 +151,7 @@ private:
 
   void updateDom(DomElement& element, bool all);
   void setTable(WTable *table);
-  void insertColumn(int column);
+  void insertColumn(int column, int row = -1);
   std::unique_ptr<WTableCell> removeColumn(int column);
 
   void undoHide();


emweb:
I'm not sure about this approach of smuggling the row number into the insertColumn() call, thereby changing the signature virtual method WTableRow::createCell().

I do wonder why that method is protected virtual yet not documented. Maybe it does not need to be. Also, it appears that WTable::createCell() does not use the row and column number by default, so it is a bit unfortunate that rowNum() gets called and then eventually the result is unused.

I think it is best that the WTableRow contains its row number (meaning that any insertions or deletions of rows require an update of the row numbers), and perhaps similarly, WTableColumn contains its column number.

Regards,
Roel

jakub-w:
I somehow missed `WTable::createCell()` was virtual. Probably because I wanted to fix it quickly and move on. This is the reason for the delay in my activity too, sorry for that. I modified the pull request as suggested.

The two tests in 989c2ad are disabled because of a bug they uncovered. I didn't want to fix it in this pull request but didn't want to remove the tests. When moving from the middle of the vector to its end an extra item is added after the inserted one, so the tests show that the vector is longer than it should be.

src/Wt/WTableRow.h
```diff
@@ -148,6 +148,7 @@ class WT_API WTableRow : public WObject
   std::string id_;
   WT_USTRING styleClass_;
   bool hidden_, hiddenChanged_, wasHidden_;
+  size_t rowNumber_;
```
emweb, line 151: This should probably be an `int` for consistency.

test/widgets/WTableTest.C
```diff
@@ -0,0 +1,269 @@
+/*
+ * Copyright (C) 2010 Emweb bvba, Kessel-Lo, Belgium.
```
emweb, line 2:
```suggestion
 * Copyright (C) 2022 Emweb bv, Herent, Belgium.
```

test/widgets/WTableTest.C
```diff
@@ -0,0 +1,269 @@
...
+
+#include <Wt/WApplication.h>
+#include <Wt/WTable.h>
+#include <Wt/Test/WTestEnvironment.h>
+
+using namespace Wt;
+
+namespace {
+Wt::Test::WTestEnvironment environment;
+Wt::WApplication testApp(environment);
```
emweb, line 16: Don't create a `WTestEnvironment` and `WApplication`. Every test should create its own, so that tests are nicely isolated.

test/widgets/WTableTest.C
```diff
@@ -0,0 +1,269 @@
...
+
+void CheckRowNumbers(WTable* table)
```
emweb, line 18:
```suggestion
void checkRowNumbers(WTable* table)
```

Wt uses `lowerCamelCase` for functions.

test/widgets/WTableTest.C
```diff
@@ -0,0 +1,269 @@
...
+{
+  for (auto i = 0; i < table->rowCount(); ++i) {
+    BOOST_REQUIRE_EQUAL(table->rowAt(i)->rowNum(), i);
+  }
+}
+void CheckColumnNumbers(WTable* table)
```
emweb, line 24:
```suggestion
void checkColumnNumbers(WTable* table)
```

test/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ IF(ENABLE_LIBWTTEST)
     utils/Base64Test.C
     utils/EraseWord.C
     utils/ParseNumber.C
+    widgets/WTableTest.C
     widgets/WSpinBoxTest.C
```
emweb, line 36:
```suggestion
    widgets/WSpinBoxTest.C
    widgets/WTableTest.C
```

T comes after S in the alphabet.

src/Wt/WTableRow.C
```diff
@@ -87,6 +88,7 @@ WTableCell *WTableRow::elementAt(int column)
 int WTableRow::rowNum() const
 {
   if (table_)
+    if (rowNumber_ != -1) return rowNumber_;
     for (unsigned i = 0; i < table_->rows_.size(); ++i)
       if (table_->rows_[i].get() == this)
 	return i;
```
emweb, line 94: In what scenario would it have to fallback to this code?

Also: this loop is not guarded by the above `if (table_)` anymore. The [readbility-misleading-indentation](https://clang.llvm.org/extra/clang-tidy/checks/readability/misleading-indentation.html) clang-tidy check catches this.

It seems to me the entire function could be simplified to `return rowNumber_;`.

src/Wt/WTable.h
```diff
@@ -197,6 +197,7 @@ class WT_API WTable : public WInteractWidget
   void expand(int row, int column, int rowSpan, int columnSpan);
   WTableCell *itemAt(int row, int column);
   void setItemAt(int row, int column, std::unique_ptr<WTableCell> cell);
+  void updateRowNumbers(int begin, int end);
```
emweb, line 200: `firstRow` and `lastRow` is probably more clear than `begin` and `end`. It's also best that you throw in some documentation that clarifies that the `lastRow` is inclusive.

test/widgets/WTableTest.C
```diff
@@ -0,0 +1,269 @@
...
+    BOOST_REQUIRE_EQUAL(table->columnAt(i)->columnNum(), i);
+  }
+}
+}
+
+BOOST_AUTO_TEST_SUITE( WTableTest )
+
+BOOST_AUTO_TEST_CASE( elementAt1 )
+{
+  auto table = cpp14::make_unique<WTable>();
```
emweb, line 36: `cpp14::make_unique` was deprecated: use `std::unique` now instead.

src/Wt/WTable.C
```diff
@@ -451,6 +458,16 @@ void WTable::moveColumn(int from, int to)
   repaint(RepaintFlag::SizeAffected);
 }
 
+void WTable::updateRowNumbers(int begin, int end) {
+  if (begin < 0 || begin >= rowCount() || end < 0 || end >= rowCount()) {
+    return;
```
emweb, line 463: If this ever happens that would be a programming error that's internal to Wt, right?

This should then either cause an alarming error message or should be omitted altogether.

src/Wt/WTable.C
```diff
@@ -451,6 +458,16 @@ void WTable::moveColumn(int from, int to)
...
+  }
+
+  for (size_t i = begin; i <= end; ++i) {
```
emweb, line 466: This loop uses `size_t`, while `begin` and `end` are `int`s?

src/Wt/WTable.C
```diff
@@ -102,7 +102,11 @@ WTableRow* WTable::insertRow(int row, std::unique_ptr<WTableRow> tableRow)
   for (auto &cell : tableRow->cells_) {
     widgetAdded(cell.get());
   }
+
+  int old_end = rows_.size();
   rows_.insert(rows_.begin() + row, std::move(tableRow));
+  updateRowNumbers(std::min(old_end, row), rowCount() - 1);
```
emweb, line 108: In what situation would this not suffice?

```suggestion
  updateRowNumbers(row, rowCount() - 1);
```

emweb (changes requested):
I've got some more comments before this could actually be merged into Wt.

PR #155, opened 2019-05-28
removed canary boost.system library after check

wrightleft:
When using the `find_package(Wt REQUIRED)` method of depending on Wt, the boost.system library's absolute path from where Wt was compiled at was found to be a dependency.

This was a result of the check added for optionally including the system component depending on the version of the Boost library. The check would set the `Boost_SYSTEM_LIBRARY` to an absolute path, and on version 1.69 of the Boost library, that absolute path would not be transformed to the import target, but would leak into the `BOOST_WT_LIBRARIES` and `BOOST_WTHTTP_LIBRARIES` variables.

Clearing the `Boost_SYSTEM_LIBRARY` variable after the check, solves this problem, though all of the variables that the check sets should be cleared. That is difficult though.


emweb:
How about doing it scoped in a function? Do you notice any problems with that?

[scoped_find_boost_system.patch.txt](https://github.com/emweb/wt/files/3280934/scoped_find_boost_system.patch.txt)

Regards,
Roel

emweb:
That last commit (42386ae) should fix it. If there's an issue anyway, feel free to comment.

Regards,
Roel

wrightleft:
The scoped function didn't do anything useful because it set the same variables in the parent scope that caused the original issue.

Clearing the variables after the check is what actually fixes the bug, and you have that
```
set(_BOOST_SYSTEM_FOUND)
set(_BOOST_VERSION)
```
So the bug is fixed, but the added function does nothing useful.

wrightleft:
Looks like that should do the trick. It makes sure that the `find_package` call can't accidentally leak anything else.

PR #156, opened 2019-07-30
Wt4:  is removeWidget() working Correct Way?

CMLDMR:
is **WWidget::removeWidget(WWidget* ptr)** remove widget by pointer.

if i set same id then removing wrong Widget.

Code Below:

```
                            for( int i = 0 ; i < 5 ; i++ )
                            {
                                auto fContainer = mFotoContainer->addWidget(cpp14::make_unique<WContainerWidget>());

                                //if enable this line, removed wrong widget in mFotoContainer
                                //fContainer->setId("fContainer");

                                fContainer->clicked().connect([=](){
                                    mFotoContainer->removeWidget(fContainer);
                                });
                            }
```


emweb:
Hello, please use our [Redmine](https://redmine.webtoolkit.eu) if you have any questions or want to report any bugs.

That being said, you're making multiple containers and giving each of them the same id. That is not allowed in HTML, so that's why you're running into problems. In general, it's not recommended to use `setId()`. Why are you using it in the first place? You'll find that most of the time you don't actually need it.

Regards,
Roel

PR #157, opened 2019-09-20
Do not write to std::cerr when opening log file successfully

mcraveiro:
Since we have opened the file successfully, it seems there isn't a lot of value in logging, and in particular logging to std::err. The logic is detailed more clearly in this ticket [1].

[1] https://redmine.emweb.be/issues/7240

### DIFF
--- a/src/Wt/WLogger.C
+++ b/src/Wt/WLogger.C
@@ -255,9 +255,6 @@ void WLogger::setFile(const std::string& path)
 #endif
   
   if (ofs->is_open()) {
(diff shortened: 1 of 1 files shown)


emweb:
I don't think we can merge this just like that. We tend to find it rather useful that the output file is printed when Wt is started.

Regards,
Roel

mcraveiro:
As discussed in the ticket, I understand that you have a lot more to worry about than my use case, so I'll close the ticket. Thanks for your time.

PR #158, opened 2019-09-22
Add WColorPicker widget

captaincrutches:
This is a simple widget that provides an `<input type="color">` element.  Its value can be retrieved or set by either a string (which must be some form of RGB) or a `WColor`.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/color for details on the element itself.


alicemargatroid:
@emweb Can we get some eyes on this or...?

PR #161, opened 2020-04-09
Implement "RESTRICT" foreign key constraint in Wt::Dbo module

ekondayan:
The RESTRICT foreign key constraint is supported by all the major databases.
It is a must have feature for quite a lot of database schemes.
It's implementation in the Wt::Dbo module is very straight forward process, so there is no apparent reason no  to do so.

Tested:
1. Wt compiles successfully 
2. SQLite3 - SQL generation for the tables works as expected

### DIFF
--- a/src/Wt/Dbo/Field.h
+++ b/src/Wt/Dbo/Field.h
@@ -17,10 +17,12 @@ namespace Wt {
 
     namespace Impl {
 const int FKNotNull = 0x01;
-const int FKOnUpdateCascade = 0x02;
-const int FKOnUpdateSetNull = 0x04;
-const int FKOnDeleteCascade = 0x08;
-const int FKOnDeleteSetNull = 0x10;
+const int FKOnUpdateCascade  = 0x02;
+const int FKOnUpdateSetNull  = 0x04;
+const int FKOnUpdateRestrict = 0x08;
+const int FKOnDeleteCascade  = 0x10;
+const int FKOnDeleteSetNull  = 0x20;
+const int FKOnDeleteRestrict = 0x40;
     }
 
 /*! \brief Type that indicates one or more foreign key constraints.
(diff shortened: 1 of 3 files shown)


emweb:
Thanks for your contribution. Your changes look good, and I know it's fairly trivial, but what would be better is if we had some unit tests, so if you can manage to write some unit tests that would be ideal. If you can check it with Sqlite3, we can make sure it works with all of the other ones.

I suppose this is a case of checking `Session::tableCreationSql()` and seeing if `Session::createTables()` works?

Regards,
Roel

ekondayan:
I've written the test, but I do not know how to include it into the framework.
Should I put into DboTest.C or should I create a new file DboTest8.C?
Could you guide me how to do it?

emweb:
That's a good start, I can take it from there.

Regards,
Roel

ekondayan:
There is something that bothers me a bit.
File Wt/Dbo/Session.C   Lines: 912
The if statement check the variable `haveSupportUpdateCascade_` which is set here 

> void Session::initSchema() const
{
...
  haveSupportUpdateCascade_ = conn->supportUpdateCascade();
...
}

Do we need to add another variable `haveSupportUpdateRestrict_` next to `haveSupportUpdateCascade_`

> void Session::initSchema() const
{
...
  haveSupportUpdateCascade_ = conn->supportUpdateCascade();
  haveSupportUpdateRestrict_ = conn->supportUpdateRestrict();
...
}

and to change the if statement to
> std::string Session::constraintString(Impl::MappingInfo *mapping, ...
{
...
  else if (field.fkConstraints() & Impl::FKOnUpdateRestrict
             && haveSupportUpdateRestrict_)
...
}

Or since all databases support the RESTRICT constraint, checking `haveSupportUpdateRestrict_` is useless?
Or to leave it like it is?

emweb:
Good observation. It appears that that check is there for Oracle. I'm still checking to see what Oracle supports. The implementation seems to suggest that `ON UPDATE` is just not supported in general, since it is also used to disable `ON UPDATE SET NULL`, so it's kind of poorly named.

In any case, I don't think it's necessary to introduce a `supportUpdateRestrict()`.

Regards,
Roel

PR #163, opened 2020-04-18
Perfect forwarding for Signal parameters

ekondayan:
This fix allows you to pass move-only objects as signal parameters. To do this, they are wrapped in std::forward(). In other words the parameters are perfectly forwarded.

An example use cases where this can be useful:

**Example emiting:**
`button->clicked().connect(
[this] { my_signal.emit(std::make_unique<Wt::WText>("Send this text widget by a signal")); });`

**Example Slot :**

`void contentLoad(std::unique_ptr<Wt::WWidget> widget)
{
  root()->addWidget(std::move(widget));
}`

### DIFF
--- a/src/Wt/Signals/signals.hpp
+++ b/src/Wt/Signals/signals.hpp
@@ -19,6 +19,7 @@
 #include <cstdint>
 #include <functional>
 #include <vector>
+#include <utility>
 
 namespace Wt { namespace Signals { namespace Impl {
 
@@ -239,7 +240,7 @@ public:
       do {
         if (link->active()) {
           try {
-            link->function(args...);
+            link->function(std::forward<Args>(args) ...);
           } catch (...) {
             link->decref();
 
@@ -327,7 +328,7 @@ struct ConnectHelper {
 			    T *target, void (V::*method)(B...))
   { 
     return signal.connect
-      ([target, method](Args... a) { (target ->* method) (a...); },
+      ([target, method](Args... a) { (target ->* method) (std::forward<Args>(a)...); },
        target);
   }
 };
@@ -368,7 +369,7 @@ struct ConnectHelper<1, Args...> {
   {
     return signal.connect
       ([target, method](B1 b1, An...) {
-	(target ->* method) (b1);
+        (target ->* method) (std::forward<B1>(b1));
       }, target);
   }
 };
@@ -390,7 +391,7 @@ struct ConnectHelper<2, Args...> {
   {
     return signal.connect
       ([target, method](B1 b1, B2 b2, An...) {
-	(target ->* method) (b1, b2);
+        (target ->* method) (std::forward<B1>(b1), std::forward<B2>(b2));
       }, target);
   }
 };
@@ -415,7 +416,7 @@ struct ConnectHelper<3, Args...> {
   {
     return signal.connect
       ([target, method](B1 b1, B2 b2, B3 b3, An...) {
-	(target ->* method) (b1, b2, b3);
+        (target ->* method) (std::forward<B1>(b1), std::forward<B2>(b2), std::forward<B3>(b3));
       }, target);
   }
 };
--- a/src/Wt/WSignal.h
+++ b/src/Wt/WSignal.h
@@ -15,6 +15,7 @@
 #endif // WT_TARGET_JAVA
 
 #include <functional>
+#include <utility>
 
 namespace Wt {
 
@@ -752,13 +753,13 @@ Wt::Signals::connection Signal<A...>
 template <class... A>
 void Signal<A...>::emit(A... args) const
 {
-  impl_.emit(args...);
+  impl_.emit(std::forward<A>(args)...);
 }
 
 template <class... A>
 void Signal<A...>::operator()(A... args) const
 {
-  emit(args...);
+  emit(std::forward<A>(args)...);
 }
 
 template <class... A>


emweb:
Hm, it's kind of strange to pass a move-only object as a signal parameter. The first slot will receive the object just fine, but the second and any subsequent slot will receive a moved from object. Do we really want to allow the possibility of these kinds of programming errors? What is the use of a signal/slot mechanism over a simple callback, if really only one slot can be connected? One thing's for sure: Wt itself will not have signals emitting move-only objects.

Here's an example of a situation where this will cause an issue:
```cpp
#include <Wt/WApplication.h>
#include <Wt/WContainerWidget.h>
#include <Wt/WPushButton.h>
#include <Wt/WText.h>

#include <memory>

class Application final : public Wt::WApplication {
public:
  Application(const Wt::WEnvironment &env);

private:
  Wt::Signal<std::unique_ptr<Wt::WWidget>> my_signal;

  void slot(std::unique_ptr<Wt::WWidget> w) {
    std::cout << std::boolalpha << (w == nullptr) << '\n';
  }
};

Application::Application(const Wt::WEnvironment &env)
  : Wt::WApplication(env)
{
    auto btn = root()->addNew<Wt::WPushButton>(Wt::utf8("Click me"));

    btn->clicked().connect([this]{
      my_signal.emit(std::make_unique<Wt::WText>("Sent over signal"));
    });

    my_signal.connect(this, &Application::slot);
    my_signal.connect(this, &Application::slot);
}

int main(int argc, char *argv[])
{
  return Wt::WRun(argc, argv, [](const Wt::WEnvironment &env) {
    return std::make_unique<Application>(env);
  });
}
```

Regards,
Roel

ekondayan:
Hi Roel,

good point, but contrary to that there are also many valid cases where this approach can be beneficial. 

Currently you can send raw pointer trough the signals/slots subsystem, which also allows you to shoot yourself in the foot. In your example if you send a raw pointer and the first slot deletes it, you will achieve the same erroneous situation. On the other side sending a move-only object is safer than sending a raw pointer, since the move-only objects are required to leave the moved object in a unspecified but a valid state. Most of the move-only objects by design implement methods for checking if the object is specified. For example the std::unique_ptr can be checked if it is a valid or a null pointer.
Another bad example is when the first slot passes the raw pointer to an object that takes the ownership and the second slot passes the raw pointer to another object that also takes the ownership. This is a recipe for disaster and is much harder to debug.

I agree with you that when the first slot acquire the ownership, the other connected slots will receive an object in a unspecified but a valid state. Sending move-only object basically enforces you to have a single slot accepting the sent signal, but it does not limit you to have a single sender, you can have multiple senders. This is basically "many to one" connection. There are many situations where this can be useful. It can be achieved with some careful planning of the signals and slots.

Use-case scenario:

I have divided my application into modules that can be easily plugged in. By a module I mean a self contained and encapsulated class that has a business logic layer and a widget factory, which generates some complex widgets.  The module itself is not aware of the outside world. The modules are created and stored in a vector in the WApplication instance. Adding the widgets to the root widget tree is handled in slots inside the WApplication. 
When the business logic layer within the module decides that it needs to show a widget, it calls the widget factory. Then the widget factory creates the widget and emit it.
The signals from the widget factory are connected to the slots in the WApplication.

```c++
class MyApplication: public Wt::WApplication
{
...
  modules.push_back(std::make_unique<MyModule>());

  for ( auto& module: modules )
  {
    module->sidebarSignal().connect(this, &MyApplication::sidebarUpdateSlot);
    
    module->contenSignal().connect(this, &MyApplication::contentUpdateSlot);
  }
...
}

void MyApplication::sidebarUpdateSlot(std::unique_ptr<Wt::WWidget> widget)
{
  ...
  //area_sidebar is a pointer to WContainerWidget, inserted in the widget tree hierarchy
  area_sidebar->clear();
  area_sidebar->addWidget(std::move(widget));
  ...
}

void MyApplication::contentUpdateSlot(std::unique_ptr<Wt::WWidget> widget)
{
  ...
  //area_content is a pointer to WContainerWidget, inserted in the widget tree hierarchy
  area_content->clear();
  area_content->addWidget(std::move(widget));
  ...
}

```


```c++
class MyModule
{
  Wt::Signal<UiChange>& sidebarSignal() { return sidebar_signal; }
  
  Wt::Signal<UiChange>& contentSignal() { return content_signal; }

  
...
  sidebar_signal.emit(widget_factory.createSidebarOne());
  sidebar_signal.emit(widget_factory.createSidebarTwo());
  sidebar_signal.emit(widget_factory.createSidebarThree());
  
  content_signal.emit(widget_factory.createContentOne());
  content_signal.emit(widget = widget_factory.createContentTwo());
  content_signal.emit(widget_factory.createContentThree());
  
...
}
```

There are countless harakiri possibilities and it is up to the programmer to take them into account. IMHO the library should not stay in the way, because that usually leads to some ugly and errorneous workarounds.

Best Regards,
Emil

emweb:
I don't know if I'm entirely convinced: if you're following the C++ core guidelines, which is what Wt does since Wt 4, a raw pointer is non-owning.

The workaround would not be to use a signal, but rather to use a simple `std::function`.

ekondayan:
I am not sure that I understand what you mean by using a std::function. May I ask you to rewrite my pseudo code above using this workaround. 

Then I can create two full examples: one will be using a std::function; the other will be using signals.

Then we can compare both examples side by side and out-weight the pros and cons of both.

emweb:
This what I quickly could do:

```c++
class MyApplication: public Wt::WApplication
{
...
  modules.push_back(std::make_unique<MyModule>());

  for ( auto& module: modules )
  {
    module->sidebarSignal() = std::bind(&MyApplication::sidebarUpdateSlot, this, std::placeholders::_1);
    
    module->contentSignal() = std::bind(&MyApplicaton::contentUpdateSlot, this, std::placeholders::_1);
  }
...
}

void MyApplication::sidebarUpdateSlot(std::unique_ptr<Wt::WWidget> widget)
{
  ...
  //area_sidebar is a pointer to WContainerWidget, inserted in the widget tree hierarchy
  area_sidebar->clear();
  area_sidebar->addWidget(std::move(widget));
  ...
}

void MyApplication::contentUpdateSlot(std::unique_ptr<Wt::WWidget> widget)
{
  ...
  //area_content is a pointer to WContainerWidget, inserted in the widget tree hierarchy
  area_content->clear();
  area_content->addWidget(std::move(widget));
  ...
}
```

```c++
class MyModule
{
  std::function<void (std::unique_ptr<Wt::WWidget>)>& sidebarSignal() { return sidebar_signal; }
  
  std::function<void (std::unique_ptr<Wt::WWidget>)>& contentSignal() { return content_signal; }

  
...
  sidebar_signal(widget_factory.createSidebarOne());
  sidebar_signal(widget_factory.createSidebarTwo());
  sidebar_signal(widget_factory.createSidebarThree());
  
  content_signal(widget_factory.createContentOne());
  content_signal(widget = widget_factory.createContentTwo());
  content_signal(widget_factory.createContentThree());
  
...
}
```

You may not want to write it exactly like this, but I hope this illustrates my point enough.

ekondayan:
```c++
#include <Wt/WServer.h>
#include <Wt/WApplication.h>
#include <Wt/WWidget.h>
#include <Wt/WPushButton.h>
```

```c++
class ModuleBind
{
   public:
    std::unique_ptr<Wt::WPushButton> createBindButton()
    {
        auto button = std::make_unique<Wt::WPushButton>("Bind");
        button->clicked().connect([this] {
          if( sidebar_changed_bind_ )
            sidebar_changed_bind_(std::make_unique<Wt::WText>("Bind Module Sidebar"));
          if( content_changed_bind_ )
            content_changed_bind_(std::make_unique<Wt::WText>("Bind Module Content"));
        });
        return button;
    }

    std::function<void(std::unique_ptr<Wt::WWidget>)>& sidebarBind() { return sidebar_changed_bind_; }
    std::function<void(std::unique_ptr<Wt::WWidget>)>& contentBind() { return content_changed_bind_; }

   private:
    std::function<void(std::unique_ptr<Wt::WWidget>)> sidebar_changed_bind_;
    std::function<void(std::unique_ptr<Wt::WWidget>)> content_changed_bind_;
};
```

```c++
class ModuleSignal
{
   public:
    std::unique_ptr<Wt::WPushButton> createSignalButton()
    {
        auto button = std::make_unique<Wt::WPushButton>("Signal");
        button->clicked().connect([this] {
            sidebar_changed_signal_.emit(std::make_unique<Wt::WText>("Signal Module Sidebar"));
            content_changed_signal_.emit(std::make_unique<Wt::WText>("Signal Module Content"));
        });
        return button;
    }

    Wt::Signal<std::unique_ptr<Wt::WWidget>>& sidebarSignal() { return sidebar_changed_signal_; };
    Wt::Signal<std::unique_ptr<Wt::WWidget>>& contentSignal() { return content_changed_signal_; };

   private:
    Wt::Signal<std::unique_ptr<Wt::WWidget>> sidebar_changed_signal_;
    Wt::Signal<std::unique_ptr<Wt::WWidget>> content_changed_signal_;
};
```

```c++
class HelloApplication : public Wt::WApplication
{
   public:
    HelloApplication(const Wt::WEnvironment& env) : Wt::WApplication(env)
    {
        setTitle("Signals vs Bind");

        auto top = root()->addWidget(std::make_unique<Wt::WContainerWidget>());
        sidebar_ = root()->addWidget(std::make_unique<Wt::WContainerWidget>());
        content_ = root()->addWidget(std::make_unique<Wt::WContainerWidget>());

        module_bind_                = std::make_unique<ModuleBind>();
        module_bind_->sidebarBind() = std::bind(&HelloApplication::sidebarUpdateSlot, this, std::placeholders::_1);
        module_bind_->contentBind() = std::bind(&HelloApplication::contentUpdateSlot, this, std::placeholders::_1);
        top->addWidget(module_bind_->createBindButton());

        module_signal_ = std::make_unique<ModuleSignal>();
        module_signal_->sidebarSignal().connect(this, &HelloApplication::sidebarUpdateSlot);
        module_signal_->contentSignal().connect(this, &HelloApplication::contentUpdateSlot);
        top->addWidget(module_signal_->createSignalButton());
    }

   private:
    void sidebarUpdateSlot(std::unique_ptr<Wt::WWidget> widget)
    {
        if( !widget ) return;
        
        sidebar_->clear();
        sidebar_->addWidget(std::move(widget));
    }
    void contentUpdateSlot(std::unique_ptr<Wt::WWidget> widget)
    {
        if( !widget ) return;
        
        content_->clear();
        content_->addWidget(std::move(widget));
    }

    Wt::WContainerWidget* sidebar_;
    Wt::WContainerWidget* content_;

    std::unique_ptr<ModuleBind> module_bind_;
    std::unique_ptr<ModuleSignal> module_signal_;
```

```c++
auto main(int argc, char** argv) -> int
{
    (void)argc;
    (void)argv;

    return Wt::WRun(argc, argv, [](const Wt::WEnvironment& env) { return std::make_unique<HelloApplication>(env); });
}
```

At first glance they seem quite equal.

# Bind Solution

If you forget to bind a function and you call it, a `std::bad_function_call` exception will be thrown.

How to mitigate it:
* Before each function call, test if it holds a callable object. In a big project this can became cumbersome and one can easily forget to do it. 

```c++
if( sidebar_changed_bind_ )
  sidebar_changed_bind_(std::make_unique<Wt::WText>("Bind Module Sidebar"));
if( content_changed_bind_ )
  content_changed_bind_(std::make_unique<Wt::WText>("Bind Module Content"));
```

# Signal Solution

If you emit a move-only object and connect two slots, the first one will take the ownership and the second one will receive the moved-from object. 

How to mitigate it:
* The programmer must ensure that there will be only one slot connected to the signals. This requirement is easily enforced even for a big project.
* The slot must place a guard in the beginning to check if the object is already moved-from. Most of the move-only object provide a method for testing if they are in a specified state. This also apply for the bind approach.

```c++
if( !widget ) return;
```

# Final toughs

| SIGNALS                                                                    | SIGNALS                                                                                                                                                                                                 | BIND                                                         | BIND                                                                                                                        |
| -------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------------- |
| PROS                                                                       | CONS                                                                                                                                                                                                    | PROS                                                         | CONS                                                                                                                        |
| the code is more expressive and is easier to understand                    |                                                                                                                                                                                                         |                                                              | not so expressive; veiled behind the  bind() and placeholders, thus it is harder to understand                              |
| emit and forget; do not care if there is a receiving slot on the other end |                                                                                                                                                                                                         |                                                              | calling unbinded function will throw an exception - before every invocation it must be tested if it holds a callable object |
|                                                                            | can be connected to more than one signal                                                                                                                                                                | binds to a single function                                   |                                                                                                                             |
|                                                                            | must check if the object is already moved-from. Since the slot is intentionally created by the programmer, it is obvious that its parameter is a move-only and the necessary precautions shall be taken | checking if the object is already moved-from can be omitted  | nevertheless is is also recommended to check it                                                                             |
|                                                                            | performance disadvantage; slower than calling a binded function                                                                                                                                         | performance advantage; faster than the signal slot subsystem |                                                                                                                             |

All that said I find the signal/slots approach more expressive and less error prone. 
The library should not tie your hands just because some careless programmer can abuse this feature. But when it is used the right way by a carefull programmer, it can lead to a better code.

emweb:
I might see some use in a maybe-move scenario:

```cpp
#include <Wt/WApplication.h>
#include <Wt/WContainerWidget.h>
#include <Wt/WPushButton.h>
#include <Wt/WText.h>

#include <memory>

class Application final : public Wt::WApplication {
public:
  Application(const Wt::WEnvironment &env);

private:
  Wt::Signal<std::unique_ptr<Wt::WWidget>&&> my_signal;

  void slot(std::unique_ptr<Wt::WWidget>&& w) {
    std::cout << std::boolalpha << (w == nullptr) << '\n';
  }
};

Application::Application(const Wt::WEnvironment &env)
  : Wt::WApplication(env)
{
    auto btn = root()->addNew<Wt::WPushButton>(Wt::utf8("Click me"));

    btn->clicked().connect([this]{
      my_signal.emit(std::make_unique<Wt::WText>("Sent over signal"));
    });

    my_signal.connect(this, &Application::slot);
    my_signal.connect(this, &Application::slot);
}

int main(int argc, char *argv[])
{
  return Wt::WRun(argc, argv, [](const Wt::WEnvironment &env) {
    return std::make_unique<Application>(env);
  });
}
```

Regards,
Roel

PR #164, opened 2020-05-12
Add traverse() method for WAbstractItemModel

ekondayan:
Add a method to traverse the whole sub tree of a valid index using the iterative  approach. The method is implemented in WAbstractItemModel but a convenience function is added to WModelIndex. For each new visited index, a callback function is called.
The callback is declared like this: std::function<bool(const Wt::WModelIndex &)>
The parameter is a const reference to the currently traversed index.
The return value indicates if the traversing should be continued or interrupted. 'true' means interrupt, 'false' means do not interrupt.
When the traversing is finished, the last traversed index is returned.


emweb:
I've personally wanted something like this, but more like an iterator, being able to go from one `WModelIndex` to the next one. Then something like this could simply be done with `std::for_each`.

~Roel

ekondayan:
What about both? 
These methods fits nicely in the WAbstractItemModel and WModelIndex interfaces, they are intuitive to use and most of the time this is all you need.

I personally would like to  have both at my disposal.

emweb:
I don't know, it doesn't really strike me as a familiar idiom. I'll have to give this some time to think about.

I think it's a bit odd that Qt never went ahead and implemented something like that for QAbstractItemModel.

~Roel

ekondayan:
Just my two cents.

Currently WModelIndex has two methods that are kind of traversal methods, they just traverse one level up and down the index tree.

```c++
WModelIndex child (int row, int column) const;
```
and
```c++
WModelIndex parent () const
```

Adding another method for deep traversal looks quite natural to me.
```c++
WModelIndex traverse(std::function<bool(const Wt::WModelIndex &)> func) const
```

ekondayan:
Hi Roel,
I've implemented the iterator. At this stage the code it is very dirty and prone to bugs but it works. 

This is how both approaches look like from the user's point of view:
```c++
auto begin_itr = WModelIndex::iterator(idx);
auto end_itr   = WModelIndex::iterator();
std::for_each(begin_itr, end_itr, [](const WModelIndex &traversed_idx) {
    if ( auto data = traversed_idx.data(ItemDataRole::Display); data.has_value() )
        std::cout << "  " << std::any_cast<Wt::WString>(data) << '\n';
    else
        std::cout << "   Item does not have Display data\n";
});
```

```c++
idx.traverse([](const WModelIndex &traversed_idx) {
    if ( auto data = traversed_idx.data(ItemDataRole::Display); data.has_value() )
        std::cout << "  " << std::any_cast<Wt::WString>(data) << '\n';
    else
        std::cout << "   Item does not have Display data\n";

    return false;
});
```

ekondayan:
Opinion needed.

Where the iterator class should reside?
1. As a nested class of WModelIndex
2. As a nested class of WAbstractItemModel

emweb:
I'd expect it to be part of `WAbstractItemModel` rather than `WModelIndex`.

I actually created an iterator of my own a while back: https://gist.github.com/RockinRoel/27ec5be4a785888c465a5121e3216a58. That one is a simple test, so not integrated into Wt in any way yet.

That one isn't ready to merge, but neither is this PR. There's a few concerns (some minor, but still necessary):

- Despite the many comments, I find the increment function hard to understand.
- I don't think it's right to name this *the* `iterator` when you're only iterating over the first column. I'd expect *the* iterator to iterate over the entire collection, not a single column.
- Nitpicking: it doesn't follow the code style of the rest of Wt. I know we don't explicitly define it somewhere, but there's a few things you'll notice if you look at the rest of Wt's code:
  - 2 spaces per indentation level. Older Wt code still turns 8 spaces into a `TAB` character, that I'm preserving for now because of `git blame`. but I'm sticking to only spaces for new code.
  - Use:
    ```
    if (condition) {
    } else {
    }
    ```
    Not:
    ```
    if( condition )
    {
    }
    else
    {
    }
    ```
  - The curly brackets of the implementation of functions does start on a new line, though.

The C++20 ranges stuff also got me thinking that maybe this could be an iterator over a range, the range being a view of a column? I don't know.

Regards,
Roel

ekondayan:
1. Improved the code readability
1.1 Added more comments
1.2 Renamed is_new to visited 
1.3 Added pastTheEnd() method to improve cede readability
1.4 Rename member variables to have more meaningful names

2. Renamed the iterator to column_iterator. 
2.1 Allow the user to choose which column is used for traversing

3. Changed the code style to your requirements

4. I haven't explored the ranges concept in details yet

Regards, 
Emil

emweb:
Sorry, I didn't mean that there wasn't sufficient documentation for your `operator++`. I just don't understand why it needs to be that complex. If it needs a lot of documentation to help understand it, that's a code smell. I don't know why your iterator needs to have five member variables either. Is there some issue I'm missing in the `operator++` that I wrote that your implementation takes care of?

The spacing around if statements is not consistent with how we usually do them either. It's usually like this: `if (condition) {`.

Regards,
Roel

ekondayan:
> Sorry, I didn't mean that there wasn't sufficient documentation for your operator++. I just don't understand why it needs to be that complex. 

This is not a viable argument.
Reading others people code is always harder than reading yours, so your statement about the code complexity is purely subjective. Neither mine nor yours is more complex than the other.
Mine code is 44 lines vs 36(equal code style and comments); 6 `if` statements vs 4; 1 cycle vs 1 cycle . That is a small difference and doesn't make my code `"that complex"`. 


> If it needs a lot of documentation to help understand it, that's a code smell.

This is not a viable argument either.
Having a lot of documentation doesn't make the code smell.


> From Wikipedia
Determining what is and is not a code smell is subjective, and varies by language, developer, and development methodology. 

> I don't know why your iterator needs to have five member variables either.

That is a good point. For performance reasons. 
* The model_ - optional. Since the iterator increment method uses the model, it needs to call WModelIndex::model() every time. When you are iterating over a very large tree(like 10,000 or 100,000 nodes), that function has to be called that many times which is a wasteful.
* The column_ - optional. The above argument also apply for this member. 
Both members can be replaced by a local variables in the increment method. That way their count can be reduced to 3(if that matters).
* start_node_ - mandatory. Preserving the start node allows you to reset the iterator from the beginning using the begin() method. It is also used to indicate the end of the traversing and prevent from escaping beyond the start node.
* lead_node_ and follow_node_ - mandatory. Used for navigating the tree.

> Is there some issue I'm missing in the operator++ that I wrote that your implementation takes care of?

Yes. 
1. Your implementation is not correct. When you start from a node that is **deeper in the tree** hierarchy and is **also a leaf** node, your algorithm escapes beyond it.
2. My iterator can be reset using the begin() method. Maybe it should be renamed to reset().

> The spacing around if statements is not consistent with how we usually do them either. It's usually like this: if (condition) {.

Will be fixed with the next commit.

emweb:
I don't think this discussion is going to resolve itself. We'll have to agree to disagree. I'm closing this PR so we don't waste eachother's time anymore.

Koen thought of the idea of adding a `begin()` and `end()` yielding an `iterator` that iterates over the entire model row per row, and creating a kind of proxy model that yields a view of a single column. I think we'll go with that approach.

Regards,
Roel

PR #165, opened 2020-06-15
Fix some GCC warnings

jwakely:


emweb:
Catching the exception by reference: sure.

But regarding the indentation: Wt used to use 2 spaces per indentation level, and every 8 spaces were converted to a TAB, which I is the Emacs default (Koen is an Emacs user).

I didn't like the mixed spaces and tabs (Wt's code looks funky in many other editors by default), so I've been using all spaces for new code, but not updating old code, mostly to make it easier to `git blame`. Maybe one day we'll have to bite the bullet and convert it to all spaces though, and remove all TABs from the code. As you've noticed, it's starting to trip up GCC's misleading indentation warning. Reverting to using TABs is going in the wrong direction, though.

Regards,
Roel

jwakely:
I just tried to make the problem lines consistent. MySQLStatement::bind_output is 70 lines long and only 5 lines use spaces for indentation, the rest uses tab. It assumed tab was preferred. I've pushed a new version replacing all tabs in those functions.

jwakely:
It doesn't actually matter if spaces and tabs are mixed, GCC will handle that. The problem was inconsistently following the "2 spaces per indentation level, and every 8 spaces were converted to a TAB" convention.

Line 825 is indented with 4 spaces, but it's inside an `if` block where the `if` itself is indented by 6 spaces. That's just wrong, and is nothing to do with tabs.

Similarly, line 777 is indented with 6 spaces, but the previous statement on line 775 is indented with 8 spaces. No tabs on either line.

emweb:
> Similarly, line 777 is indented with 6 spaces, but the previous statement on line 775 is indented with 8 spaces. No tabs on either line.

Yeah, the indentation is indeed just wrong in some places. Likely this was caused by different developers using different indentation settings and seeing the TABs differently, which is why I'm moving away from that.

Maybe GCC is indeed cool with us mixing 8 spaces and TABs... anyway

Anyway, thanks, your PR looks good.

Regards,
Roel

PR #167, opened 2020-09-16
Explicitly link to Boost Unittest Framework libraries for each test application.

chr-thien:
This change allows using Boost 'tagged' layout (autolink disabled).
I tested with CMake 3.18.2 and Boost 1.74.


emweb:
Could you comment on in which situation (configuration, platform) this change becomes relevant? I didn't find any linking problems in any configuration I've tested so far.

Regards,
Roel

emweb:
Nevermind, I got it. I fixed it in another commit in a slightly different way.

PR #169, opened 2020-12-05
Update WCartesianChart.C

jeroen-zpam:
fix memory leak


emweb:
While a valid fix for Wt 3 (and I would have merged it at the time), we've discontinued Wt 3, so it will no longer be merged.

Regards,
Roel

PR #172, opened 2020-12-15
3.7 release

jeroen-zpam:
added configuration option whether to add the "X-Frame-Options", "SAMEORIGIN" header


MatthiasVC:
Hello Jeroen,

I am very sorry this PR was left open and unattended to for so long.
This will be part of Wt 4.11.2, as per the [ticket](https://redmine.emweb.be/issues/6584).

Since we also have a special case for JWt for this, I prefer to handle this internally. 
If you do still read this update, I appreciate what you did, and I regret this has remained unhandled for so long.

Again, my apologies.

PR #173, opened 2021-01-06
The demo code was no longer valid referring to WSocketNotifier::Read

JanKovis:
but with enum class advent the valid reference is WSocketNotifier::Type::Read

pure documentation improvement


emweb:
Thanks!

Roel

PR #175, opened 2021-01-26
Fixed '[c|tr].size is not a function' error when using jQuery 3.x.

skoehler47:
While using jQuery 3.x with WApplication::requireJQuery, the following error occurs if the server is restarted while there are still active sessions, thus forcing the sessions to reload.

jquery-size

This is caused by the use of the jQuery size() function, which was deprecated in jQuery 1.8 and removed in jQuery 3.0. Replacing the size() function with the length property as suggested in https://api.jquery.com/size/ fixes the issue, while remaining full compatibility with jQuery 1.0+.


emweb:
What does the c refer to in the title? Is that the error you got with the minified version?

skoehler47:
That's right. In the non-minified version the call is tr.size(); in the minified version it is c.size().

emweb:
Whoops, I managed to push it in a way that does not mark this as merged.

It's fixed in commit f8869aedd1266c67ff58f87da099953dc64ccadf. I just changed the commit message to include your rationale.

PR #177, opened 2021-01-26
Fixed several issues in Wt::Dbo::collection<>

skoehler47:
This pull request covers/fixes several issues with Wt::Dbo::collection<>:

- The collection::iterator operator == may cause an access violation due to a missing null-check in the third case; this can occur if impl_ == nullptr && other.impl_ != nullptr && other.impl_.ended_ == false, which is not covered by the first 2 cases.
- collection::iterator and collection::const_iterator are derived from std::iterator, which is deprecated in C++17. Since deriving from std::iterator does nothing but adding some type definitions, adding those typedefs directly to iterator/const_iterator solves the deprecation issue while being fully compatible with C++14 and before.
- const_iterator declares the assignment operator operator = (iterator), but the definition is missing. I added one which mimics the behavior of the const_iterator(iterator) constructor.
- There were some 32/64bit signed/unsigned type conversion issues in collection::size() and count() functions. I added some static_casts to resolve the numerous warnings caused by this, though i suggest a more thorough fix for the issue

### DIFF
--- a/src/Wt/Dbo/collection.h
+++ b/src/Wt/Dbo/collection.h
@@ -122,9 +122,16 @@ namespace Wt {
 
     /*! \brief Iterator.
      */
-    class iterator : public std::iterator<std::input_iterator_tag, C>
+    class iterator
     {
     public:
+
+      typedef std::input_iterator_tag iterator_category;
+      typedef                       C value_type;
+      typedef          std::ptrdiff_t difference_type;
+      typedef             value_type *pointer;
+      typedef             value_type& reference;
+
       /*! \brief Copy constructor.
        */
       iterator(const iterator& other);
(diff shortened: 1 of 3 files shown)


emweb (changes requested):
Could you also reword your commits so that they have a short title and a longer description (with line breaks), instead of having the entire commit message on a single long line?

skoehler47:
Done. I hope it is more to your liking now.

I usually prefer 'using' over 'typedef' too, but decided to go with typedef in this case to keep consistency with the rest of Wt::Dbo.

emweb:
Yeah, I suppose the rest of Dbo is still in the old style, but I think for new code we can go ahead and move towards the new style.

I'll make some modifications on your branch. There are also some simple tests I want to add.

Regards,
Roel

PR #180, opened 2021-08-20
Fixes for invalid MySQL timestamps

skoehler47:
When examining a MySQL/MariaDB database I use in a Wt project I noticed that all timestamps, without a single exception, had the format `YYYY-MM-DD HH:MM:SS.000` - the millisecond part was always exactly zero, although the original timestamps definitely had non-zero millisecond parts. The issue occurred when storing `WDateTime` as well as `std::chrono::system_clock::time_point`.

Some digging later I found the culprit in Wt's MySQL backend.  In `MySQL.C` the `ts->second_part` value is initialized with the millisecond value. According to MySQL documentation (see [MYSQL_TIME Struct Reference](https://dev.mysql.com/doc/dev/mysql-server/latest/structMYSQL__TIME.html)) that field contains a microsecond value. A little fiddling with the datetime precision revealed that the value was indeed stored a microseconds, producing outputs like `HH::MM::SS.000123`. I could fix that issue by converting the milliseconds to microseconds and write that value to `ts->second_part`. 

I also fixed the type casting. The previous version performed the cast first and calculated the remainder second, which leads to incorrect results. Minified example: Storing the remainder of dividing 1234 by 100 in a uint8_t would first cast 1234 to uint8_t (= 210) and then calculate the wrong remainder 10 - instead of the correct value 34. Dividing first produces the expected result.

Reading timestamps from database produced a similar issue. The supposed millisecond part of the timestamp is returned as microseconds. `HH:MM:SS.123` gets a `ts->second_part` value of 123000, which was interpreted as 123000 milliseconds in `MySQL.C` - resulting in timestamps being off up to 1000 seconds (~17 minutes) from the actual ones stored in the database. Fixed that too.

Last change is a performance improvement: I added two `#ifdef` blocks to prevent the (quite expensive) superfluous creation of debug information in the cases that no debug messages are logged, aka cases where `LOG_DEBUG` macro is defined as empty.

I verified the fixes with my local database - timestamps work as expected now. I also checked the `std::chrono::duration` functionality in the MySQL backend, which already perform the proper microsecond casts.

Best regards
Steven

### DIFF
--- a/src/Wt/Dbo/backend/MySQL.C
+++ b/src/Wt/Dbo/backend/MySQL.C
@@ -281,9 +281,11 @@ class MySQLStatement final : public SqlStatement
 
       std::time_t t = std::chrono::system_clock::to_time_t(value);
       std::tm *tm = thread_local_gmtime(&t);
+#ifdef WT_DEBUG_ENABLED
       char mbstr[100];
       std::strftime(mbstr, sizeof(mbstr), "%Y-%b-%d %H:%M:%S", tm);
       LOG_DEBUG(this << " bind " << column << " " << mbstr);
+#endif
 
       MYSQL_TIME*  ts = (MYSQL_TIME*)malloc(sizeof(MYSQL_TIME));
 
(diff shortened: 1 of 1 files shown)


emweb:
This has been merged through the Redmine issue.

PR #181, opened 2021-08-26
Fixed leaks on WFileDropWidget (as well as temporary files that are not being destroyed)

epoupon:
Spotted it using valgrind, and I also noticed that temporary files were not deleted


epoupon:
@emweb No news on this?

dkopecek:
This patch is no longer needed since the 4.7.0 release, is that correct? The internal vector holding `File`s now uses `unique_ptr`.

epoupon:
> This patch is no longer needed since the 4.7.0 release, is that correct? The internal vector holding `File`s now uses `unique_ptr`.

Yes indeed

PR #183, opened 2021-10-31
Fixes for WT-9270: 'Unique IDs' are not unique

skoehler47:
This pull requests belongs to [Issue 9270](https://redmine.webtoolkit.eu/issues/9270), which contains details and further informations about the issue.

### DIFF
--- a/src/Wt/WJavaScriptSlot.C
+++ b/src/Wt/WJavaScriptSlot.C
@@ -11,7 +11,11 @@
 
 namespace Wt {
 
-int JSlot::nextFid_ = 0;
+#ifdef WT_THREADED
+  std::atomic<unsigned> JSlot::nextFid_ = 0;
+#else
+  unsigned JSlot::nextFid_ = 0;
+#endif // WT_THREADED
 
 class WStatelessSlotImpl : public WStatelessSlot {
 public:
--- a/src/web/DomElement.C
+++ b/src/web/DomElement.C
@@ -153,7 +153,11 @@ static_assert(sizeof(defaultInline_) / sizeof(defaultInline_[0]) == static_cast<
 
 namespace Wt {
 
-int DomElement::nextId_ = 0;
+#ifdef WT_THREADED
+  std::atomic<unsigned> DomElement::nextId_ = 0;
+#else
+  unsigned DomElement::nextId_ = 0;
+#endif // WT_THREADED
 
 DomElement *DomElement::createNew(DomElementType type)
 {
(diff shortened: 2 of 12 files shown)


src/Wt/WJavaScriptSlot.C
```diff
@@ -11,7 +11,7 @@
 
 namespace Wt {
 
-int JSlot::nextFid_ = 0;
+std::atomic<unsigned> JSlot::nextFid_ = 0;
```
emweb, line 14: This should be non-atomic depending on `WT_THREADED`. Same comment for all the other ones.

src/web/DomElement.C
```diff
@@ -734,7 +734,7 @@ void DomElement::setJavaScriptEvent(EscapeOStream& out,
   // events on the dom root container are events received by the whole
   // document when no element has focus
 
-  int fid = nextId_++;
+  unsigned fid = nextId_++;
```
emweb, line 737: So all of them use `.fetch_add(1, std::memory_order_relaxed)` except for this one...

src/web/DomElement.C
```diff
@@ -1123,7 +1123,7 @@ std::string DomElement::createVar() const
 {
 #ifndef WT_TARGET_JAVA
   char buf[20];
-  std::sprintf(buf, "j%d", nextId_++);
+  std::sprintf(buf, "j%u", nextId_++);
```
emweb, line 1126: ...and this one?

This would use `std::memory_order_seq_cst`, which theoretically has more overhead. I experimented a bit and the difference seems to be negligible in practice. On AMD64 it would generate the same assembly code, on AArch64 performance was very close. So I would be fine with just all of them using `operator++` instead of `fetch_add`.

emweb (changes requested):
Thanks for your PR and your thorough analysis.

I only have a few small remarks.

Regards,
Roel

src/web/DomElement.C
skoehler47: That's actually an oversight on my end...

src/web/DomElement.C
skoehler47: No objections against using `operator++` on my part. This actually makes switching between the atomic and the not-atomic version much easier.

skoehler47:
I updated the commits with some `#ifdef` blocks to distinguish between the threaded and the non-threaded version. The threaded version uses `std::atomic<unsigned>` for its IDs, the non-threaded version uses `unsigned`. I also replaced all `fetch_all` occurrences with `operator++`, which allows incrementing and using the IDs the same way in both versions.

emweb (approved):
Looks good to me :slightly_smiling_face: 

Thanks,
Roel

PR #185, opened 2022-01-30
replace boost::filesystem with std::filesystem

xth:
replace boost::filesystem with std::filesystem
set CMAKE_CXX_STANDARD to 17


emweb:
Sorry, I don't think we're updating Wt to C++17 quite yet.

The way we've handled it with `std::any` and `std::chrono`'s date/timezone support is by adding a CMake flag for it. We could consider that, dropping the dependency on Boost Filesystem if the used standard is C++17.

Regards,
Roel

emweb:
Closing this for the reasons I stated: we're currently still keeping C++14 compatibility.

PR #186, opened 2022-02-04
Fix WAxisSliderWidget and chart viewport fail to sync

Apivan:


Apivan:
Fix issue  https://redmine.webtoolkit.eu/issues/9719

src/js/WCartesianChart.js
```diff
@@ -512,8 +512,8 @@ WT_DECLARE_WT_MEMBER_BIG
         for (i = 0; i < len(sliders()); ++i) {
           var o = $('#' + sliders()[i]);
           if (o) {
-            var sobj = o.data('sobj');
-            if (sobj && sobj.xAxis === xAx) {
+            var sobj = o[0].WtSObj;
```
RockinRoel, line 515: Typo? This should be `wtSObj`.

In combination with the other remark, this would become:

```cpp
var sobj = o.wtSObj;
```

src/js/WCartesianChart.js
```diff
@@ -512,8 +512,8 @@ WT_DECLARE_WT_MEMBER_BIG
         for (i = 0; i < len(sliders()); ++i) {
           var o = $('#' + sliders()[i]);
```
RockinRoel, line 513: I'd prefer it in the non-jQuery way here (we want to get rid of our jQuery dependency eventually):

```cpp
var o = document.getElementById(sliders()[i]);
```

RockinRoel (changes requested):
Thanks for you contribution. How did you minify the source code? I think it's likely not the same minifier as we're using. I'll do the minification myself (eventually we want to switch all of Wt's JavaScript over to another minifier).

Regards,
Roel

Apivan:
I used babel node js to minify.

emweb:
Yeah, we're using an old version of Google Closure compiler at the moment. I minified it and made another small fix (`o.wtSObj` instead of `o[0].wtSObj`) and did a manual merge, so this can be closed now.

Regards,
Roel

PR #189, opened 2022-05-13
Added support for long in json

felix642:
I don't see why JSON does not support long values.
I added a constructor that supports long values and the associated code.

### DIFF
--- a/src/Wt/Json/Value.C
+++ b/src/Wt/Json/Value.C
@@ -278,22 +278,6 @@ Value::operator int() const
     throw TypeException(type(), Type::Number);
 }
 
-Value::operator long long() const
-{
-  const std::type_info& t = v_.type();
-
-  if (t == typeid(double))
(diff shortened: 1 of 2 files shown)


notorand-it:
... and long long too?

emweb (changes requested):
Thanks for your contribution. I have some minor remarks before I can merge it.

As for why the JSON library may lack certain features: the Wt JSON library was created more out of necessity than as a feature of Wt. Wt::Auth's OAuth support needed to use JSON, so we added a JSON library back in 2011. We didn't have the wealth of JSON libraries that we have now, like [JSON for Modern C++](https://json.nlohmann.me/), [Boost.JSON}(https://www.boost.org/doc/libs/1_79_0/libs/json/doc/html/index.html), or [RapidJSON](https://rapidjson.org/).

Regards,
Roel

PR #191, opened 2022-06-08
Add information for using WResource

limitedAtonement:
I didn't see any other documentation that lays this pointer out, so I think this is the best place for it.

It might also be helpful to point to a pertinent example such as `blog.C:49` or the URL Params example (lines 53 and following).

### DIFF
--- a/src/Wt/WResource.h
+++ b/src/Wt/WResource.h
@@ -86,6 +86,10 @@ enum class ContentDisposition {
  * \endcode
  * \endif
  *
+ * In order to configure your web server to serve such a resource, either
+ * create a link to it using WAnchor or WImage, or add it to your WServer
+ * using WServer::handleRequest().
+ *
  * <h3>Concurrency issues</h3>
  *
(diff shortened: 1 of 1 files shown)


limitedAtonement:
Please let me know if something should be changed here to make it suitable for inclusion in the project, or if you reject the pull request, I'll be happy to close.

src/Wt/WResource.h
```diff
@@ -86,6 +86,10 @@ enum class ContentDisposition {
  * \endcode
  * \endif
  *
+ * In order to configure your web server to serve such a resource, either
+ * create a link to it using WAnchor or WImage, or add it to your WServer
+ * using WServer::handleRequest().
```
RockinRoel, line 91: I assume you mean `addResource`, not `handleRequest`.

Also: The "global and private resources" heading already makes mention of it. The only thing that's lacking is how to serve a private resource (i.e. through a `WLink`).

limitedAtonement:
I think you're right: this is adequately addressed already. Thank you!

PR #193, opened 2022-06-12
Fixed a possible access violation in WebSession

skoehler47:
This pull request contains a fix for a segmentation fault during application creation, if the application create function used for an entry point returns a nullptr - as described in [Issue 10483](https://redmine.webtoolkit.eu/issues/10483).

### DIFF
--- a/src/web/WebSession.C
+++ b/src/web/WebSession.C
@@ -680,9 +680,15 @@ bool WebSession::start(WebResponse *response)
 {
   try {
     app_ = controller_->doCreateApplication(this).release();
-    if (!app_->internalPathValid_)
-      if (response->responseType() == WebResponse::ResponseType::Page)
-        response->setStatus(404);
+    if (app_) {
+      if (!app_->internalPathValid_) {
+        if (response->responseType() == WebResponse::ResponseType::Page) {
+          response->setStatus(404);
+        }
+      }
+    } else {
+      throw WException("WebSession::start: ApplicationCreator returned a nullptr");
+    }
   } catch (std::exception& e) {
     app_ = nullptr;
 


src/web/WebSession.C
```diff
@@ -680,9 +680,16 @@ bool WebSession::start(WebResponse *response)
...
-    if (!app_->internalPathValid_)
-      if (response->responseType() == WebResponse::ResponseType::Page)
-        response->setStatus(404);
+    if (app_) {
+      if (!app_->internalPathValid_) {
+        if (response->responseType() == WebResponse::ResponseType::Page) {
+          response->setStatus(404);
+        }
+      }
+    }
```
emweb, line 689: Small style comment: in the Wt source code we put the `else` on the same line as the closing bracket, so like this:

```cpp
} else {
```

src/web/WebSession.C
```diff
@@ -680,9 +680,16 @@ bool WebSession::start(WebResponse *response)
...
+    else {
+      throw WException("WebSession::start: doCreateApplication returned a nullptr");
```
emweb, line 691: I think it would be better to refer to publicly documented `ApplicationCreator`, since `doCreateApplication` is an implementation detail and we can't expect users to be familiar with it.

emweb (changes requested):
Thanks for your contribution. I have a few small remarks. If you can address them that would be great :slightly_smiling_face: 

Roel

src/web/WebSession.C
skoehler47: Hi Roel,

you're welcome. I've just updated the commit according to your remarks.

Steven

PR #194, opened 2022-06-23
Added missing timeout default initialization in FixedSqlConnectionPool

skoehler47:
This pull request contains a fix for the missing timeout default initialization in `FixedSqlConnectionPool` - as described in [Issue 10543](https://redmine.webtoolkit.eu/issues/10543).

### DIFF
--- a/src/Wt/Dbo/FixedSqlConnectionPool.C
+++ b/src/Wt/Dbo/FixedSqlConnectionPool.C
@@ -29,7 +29,7 @@ struct FixedSqlConnectionPool::Impl {
   std::condition_variable connectionAvailable;
 #endif // WT_THREADED
 
-  std::chrono::steady_clock::duration timeout;
+  std::chrono::steady_clock::duration timeout{ std::chrono::steady_clock::duration::zero() };
   std::vector<std::unique_ptr<SqlConnection>> freeList;
 };
 


src/Wt/Dbo/FixedSqlConnectionPool.C
```diff
@@ -39,6 +39,7 @@ FixedSqlConnectionPool::FixedSqlConnectionPool(std::unique_ptr<SqlConnection> co
 {
   SqlConnection *conn = connection.get();
   impl_->freeList.push_back(std::move(connection));
+  impl_->timeout = std::chrono::steady_clock::duration::zero();
```
emweb, line 42: I think it's preferable to default-initialize the `timeout`, like:

```cpp
std::chrono::steady_clock::duration timeout{};
```

emweb (changes requested):

src/Wt/Dbo/FixedSqlConnectionPool.C
skoehler47: I'd usually agree, but in this case I decided to go with the `zero` initialization since you explicitly compare against `zero` in `getConnection` which implied to me that you expect the default value to be `zero`, and not whatever aggregate initialization delivers (though it should be 0 too). I took a quick peek in Visual Studio's STL implementation and they explicitly initialize the `zero` function return value with 0, so even they don't rely on aggregate initialization for the `zero` duration.

I can of course change it if you want. Just let me know. In the end, what really matters is that the issue gets fixed. Otherwise I might offer to reword the commit message to make clear that it is a zero and not a default initialization, to ease possible confusion.

RockinRoel: I consulted the standards document and now I'm indeed doubting whether it is indeed mandated that `std::chrono::steady_clock::duration::zero()` is the same as `std::chrono::steady_clock::duration{}`.

I guess to be completely sure it's fine we should use `zero()`. I'd still move it to the declaration of the `struct` rather than here, though. In a lot of cases in Wt we can't do it because our Wt-to-JWt toolchain doesn't support it, but it's fine in Wt::Dbo.

emweb (changes requested):

src/Wt/Dbo/FixedSqlConnectionPool.C
```diff
@@ -24,6 +24,11 @@ namespace Wt {
 LOGGER("Dbo.FixedSqlConnectionPool");
 
 struct FixedSqlConnectionPool::Impl {
+
+  Impl()
+    : timeout(std::chrono::steady_clock::duration::zero())
+  { /* empty */ }
```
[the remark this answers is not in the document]
skoehler47: Sorry, my mistake. Done.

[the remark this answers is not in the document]
emweb: Thanks :slightly_smiling_face:

PR #195, opened 2022-07-07
Fixed a possible access violation in WApplication::UpdateLock

skoehler47:
This pull request contains a fix for a possible access violation in `WApplication::UpdateLock` - as described in [Issue 10614](https://redmine.webtoolkit.eu/issues/10614).

### DIFF
--- a/src/Wt/WApplication.C
+++ b/src/Wt/WApplication.C
@@ -1470,6 +1470,14 @@ WApplication::UpdateLock::UpdateLock(WApplication *app)
 #ifndef WT_THREADED
   return;
 #else
+
+  // check if 'app' is a nullptr to prevent an access violation
+  // in the following code
+  if (!app) {
+    ok_ = false;
+    return;
+  }
+
   /*
    * If we are already handling this application, then we already have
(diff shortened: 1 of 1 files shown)


src/Wt/WApplication.C
```diff
@@ -1498,8 +1506,12 @@ WApplication::UpdateLock::UpdateLock(WApplication *app)
   WebSession::Handler *handler = WebSession::Handler::instance();
 
   createdHandler_ = false;
-  if (handler && handler->haveLock() && handler->session() == app->session_)
+  if (!app) {
+    return;
+  }
```
emweb, line 1511: Can you remove this bit? We only need the first diff chunk. This chunk is used in JWt, which already does the right thing because this constructor is `private` and the lock can only be obtained with `getUpdateLock()`.

Regards,
Roel

emweb (changes requested):

src/Wt/WApplication.C
skoehler47: Sure thing. Done. 

Regards,
Steven

emweb:
Thanks!

Roel

PR #196, opened 2022-12-13
Touch up documentation on Wt::WSignal

limitedAtonement:
The documentation was missing qualifications concerning under which conditions the signal would NOT be disconnected. That is, for instance if you pass a lambda function object (which doesn't inherit from `Wt::Core::observable`), `Wt::WSignal` has no way to unregister or disconnect the signal.

### DIFF
--- a/src/Wt/WSignal.h
+++ b/src/Wt/WSignal.h
@@ -166,8 +166,10 @@ private:
  * the same parameters in the member function, or less (leaving out
  * parameters at the end).
  *
- * The signal automatically disconnects from the slot when the
- * target is deleted. In addition, the signal may be deleted at any
+ * When the receiver function is an object method, the signal will
+ * automatically be disconnected when the object is deleted, as long
+ * as the object inherits from WObject (or Wt::Core::observable).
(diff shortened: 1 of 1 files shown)


RockinRoel:
`connect()` always returns `connection` object, on which you can call `disconnect()`, so you can always disconnect a slot.

The slot isn't really automatically disconnected when the `observable` goes away, so it's inaccurate to say so. What really happens is that the slot is wrapped in a function that checks whether the `observing_ptr` is valid before executing the slot. As it is right now, the slot will not be cleaned up; it simply won't do anything.

limitedAtonement:
Thank you for your pointers! I just noticed this undocumented `Wt::Signals::connection` when perusing the source. I verified that the return value is mentioned in the docs! (https://www.webtoolkit.eu/wt/doc/reference/html/classWt_1_1Signal.html#aed69704ae147dbb551c3074c8214940c) Thank you for the pointer!

RockinRoel:
That `connection` object should probably be documented properly, yeah.

PR #199, opened 2023-01-10
Fix WTreeView loading. See #11240

jhnstrk:
When rows are removed, `WTreeView::modelRowsAboutToBeRemoved` calls `isExpandedRecursive` to determine if rendered tree nodes are impacted by the removal. However, `isExpandedRecursive` was always returning false for the root node. This seems odd, when the root node is always expanded. This behaviour caused a bug because the `removedHeight_` was not being set. This lead to the `childrenHeight` not being updated, and the loading isn't triggered.

Switching  `isExpandedRecursive`  to true for the root index fixes the issue ... but maybe it was intentionally set like this, but I can't see why.

### DIFF
--- a/src/Wt/WTreeView.C
+++ b/src/Wt/WTreeView.C
@@ -1739,7 +1739,7 @@ bool WTreeView::isExpandedRecursive(const WModelIndex& index) const
     if (index != rootIndex())
       return isExpanded(index.parent());
     else
-      return false;
+      return true;
   } else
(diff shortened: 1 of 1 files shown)


emweb:
Thanks. Having it return `false` on the root index doesn't make any sense to me either.

I manually merged this with some changes (adding your description to the commit message and adding release notes), so I'm closing this PR.

Roel

PR #200, opened 2023-02-14
Support for C++20 u8string

Apivan:
Since WString already stored utf8_ as std::string. This is just additional interfaces for accepting std::u8string arguments. No conversion needed.

image
https://en.cppreference.com/w/cpp/language/types#char8_t

### DIFF
--- a/src/Wt/WString.C
+++ b/src/Wt/WString.C
@@ -67,6 +67,13 @@ WString::WString(const std::u32string &value)
   utf8_ = Wt::toUTF8(value);
 }
 
+#ifdef __cpp_char8_t
+WString::WString(const std::u8string& value)
+{
+  utf8_ = std::string(value.begin(), value.end());
+}
+#endif
+
 WString::WString(const char *value, CharEncoding encoding)
   : impl_(nullptr)
 {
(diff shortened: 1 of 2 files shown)


emweb:
Thanks for your contribution.

Ideally Wt would use `u8string` as its internal representation and use it throughout, but maybe that's more of a Wt 5 thing (we don't have plans for another major release anytime soon).

I have a few remarks:

- Binary compatibility: if you're using C++20, but Wt was compiled with C++17, this will cause linker errors when the `u8string` functions are used. I suggest perhaps we make the `char8_t`/`std::u8string` functions `inline`. We do already have an accidental incompatibility due to the `date` library using `std::string_view` instead of `std::string` with C++17, but let's try not to introduce new incompatibilities.
- We need unit tests for these functions.
- The `operator std::u8string()` function is not implemented correctly, consider localized `WString`s or `WString`s with arguments. You'd have to construct a `u8string` from the result of `toUTF8()`.
- The `#ifdef`s should probably contain the documentation. I think if we want to generate proper documentation, we'll need to check `#if defined(__cpp_char8_t) || defined(DOXYGEN_ONLY)`. Use the `doxygen` target to verify that the documentation is generated properly.
- The `WString(const std::u8string&)` constructor lacks a doc comment.

Regards,
Roel

Apivan:
Thank you for your reply. I believe that using std::u8string internally in the future Wt would be a better solution. Initially, I attempted to migrate our project source code to u8string in order to solve some issues, but I then realized that most third-party libraries do not support this. Changing large project to u8string would be a massive undertaking and cause excessive copying and conversion . So I withdraw this pull request.

PR #201, opened 2023-02-20
WCssDecorationStyle: use constructor delegation to initialize cursor_ in copy constructor

HugoLopata:
WCssDecorationStyle::cursor_ was used uninitialized in copy constructor via copy - setCursor - if 
pointed by valgrind


emweb:
Sadly, we can't do it that way, since our Wt-to-JWt translation toolchain doesn't support delegating constructors. We could initialize it to `Cursor::Auto`, though.

HugoLopata:
Ok, I replaced constructor delegation with full initializer list from main constructor. It is safer to initialize all member variables.

PR #204, opened 2023-07-23
Option added to specify Cache-Control header for static files.

m7913d:


MatthiasVC:
Thank you,

could you please also update the documentation for this, so that people are aware of the possibility to use this.

There'd need to be entries in:
- INSTALL
- INSTALL.html
- INSTALL.win32.html
- doc/main

I believe the best fit is under `HTTP/WebSocket server options` in section `10.3`.

I you prefer to leave documentation to us, you can let me know.

m7913d:
Documentation added. However, I added it under general as the option is used for both http and https. Also the related docroot option is documented under general.

Feel free to update/move the documentation as appropriate.

MatthiasVC:
Thank you for your submission!

I apologize for the wait. I prefer to do a bunch of these at once to avoid our internal origin to get a messier history (it has a different history from this project).

PR #205, opened 2023-07-28
Handle enableAjax for WCarthesianChart when progressive bootstrap is enabled

cwitting:
By registering the event handlers on the enableAjax call, the interactive components of WCarthesianChart also works when progressive bootstrap is enabled.

This has not been tested in depth, but seems to fix the issue we had with interactive plots not working as described in:
https://redmine.emweb.be/issues/7514

But maybe there is a better way to fix it?

### DIFF
--- a/src/Wt/Chart/WCartesianChart.C
+++ b/src/Wt/Chart/WCartesianChart.C
@@ -1395,6 +1395,7 @@ WCartesianChart::WCartesianChart()
     onDemandLoadingEnabled_(false),
     loadingBackground_(StandardColor::LightGray),
     cObjCreated_(false),
+    hasAjax_(false),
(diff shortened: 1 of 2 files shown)


MatthiasVC:
Hello,

if I apply this change, and define a `wt_config.xml` with `progressive-bootstrap` set to `true`, the example charts still do not function.

As indicated on the ticket, it's likely best to work around this issue by defining `bootstrap-method` for specific paths, disabling it on paths with charts (as can be seen in [this example](https://github.com/emweb/wt/blob/master/examples/widgetgallery/approot/wt_config.xml))

We are aware that there are issue with the bootstrap approach, but they span over a wide array of widgets. Likely something more structural will have to change there.

PR #210, opened 2024-02-16
Fix https://redmine.emweb.be/issues/12424

Apivan:


MatthiasVC:
Hello @Apivan,

again, sorry for the huge delay. We believe that this was resolved as part of https://redmine.emweb.be/issues/12366.

I'll leave this open for a while, in case it was not.

PR #211, opened 2024-02-16
Fix doJavaScript("") crash

Apivan:
minor problem: calling doJavaScript({}); crashes


Apivan:
Duplicated with redmine issues

PR #212, opened 2024-02-21
Do Install pdb(s) for msvc

Apivan:


MatthiasVC:
Hello Apivan,

I created [WT-13700](https://redmine.emweb.be/issues/13700) to track this, as I am going to do a small change to it (namely guarding the PDB target behind a shared build as well).

This will pass the internal review process, and thus be tracked internally.
I will take your original commit, to allow for credit to remain, and apply that internally.

You will see this appear soon on the master branch.

Best,
Matthias

PR #213, opened 2024-02-23
Fix  test.wt  build error from missing openssl header

Apivan:
The header is included by \test\utils\SslUtils.C


MatthiasVC:
Hi @Apivan,

sorry for the very late reply. This was noticed during the implementation of https://redmine.emweb.be/issues/12461.
And has been fixed as part of that MR.

Thank you for the PR!

PR #214, opened 2024-04-10
Fix the logical judgment of ENABLE_LIBWTTEST

MonicaLiu0311:
Fix the logical judgment of [ENABLE_LIBWTTEST](https://github.com/emweb/wt/blob/4.10.4/src/CMakeLists.txt#L563-L612):
```
IF(ENABLE_LIBWTTEST)
  ...
ELSE(ENABLE_LIBWTTEST)
  MESSAGE("** Not building libwttest")
ENDIF(ENABLE_LIBWTTEST)
```


MonicaLiu0311:
I misunderstood, and it’s okay to judge this way.

PR #217, opened 2024-08-17
web/FileUtils.C: Include boost/filesystem/directory.hpp explicitly

erikbs:
As of Boost 1.85, including boost/filesystem/operations.hpp no longer implicitly includes boost/filesystem/directory.hpp


MatthiasVC:
Hi,

similar to #219  this was performed internally already, and will be closed.

Thank you for your contribution.

PR #218, opened 2024-08-17
Server.C: use separate variable for IPv6 resolver query

erikbs:
Boost 1.85 no longer provides assignment operators for asio::ip::basic_resolver_query. This is because they recently added copy/move constructors, so the assignment operators are now implicitly deleted. I am not sure if this is intentional, as they do not follow the rule of five in this class. Either way Wt fails to build when attempting to reuse a query variable. The error can be avoided by creating a separate variable for the IPv6 query instead of reusing the one for IPv4 (the compiler will probably optimise the code anyway).


MatthiasVC:
Hi @erikbs ,

apologies for the delay. Much like the other PR, this was resolved as part of https://redmine.emweb.be/issues/12676.

Thank you for the PR, and I again do apologize for the delay.

PR #220, opened 2024-09-17
Fix popups going off screen vertically (upwards)

tcld:
Imagine a popup menu being 60% as tall as your screen real estate. If you open this popup in the (vertical) center of the screen, the current algorithm to place it will recognize that there isn't enough space _underneath_ the click position to display it (50% + 60% > 100%).
The popup is then shown _above_ the click position without any further checks, despite there not being enough space available there, either. (50% - 60% < 0%)

This PR slightly modifies this behavior by adding a second check: If there isn't enough space above the click position, the menu is moved down as far as necessary to display it. (At this point in the code we know that the menu is small enough to fit on the screen, so we only have to consider where - not whether - to place the menu.)

This change affects all popup menus (e.g. on buttons and comboboxes), but it doesn't fully fix this issue when observed on comboboxes, since they might be parented in more complex ways, which affects where their popup-menus can be shown.


MatthiasVC:
Hello Stephan,

you may have gotten the update that the [ticket](https://redmine.emweb.be/issues/13075) has been merged (internally, soon to be added to Github).
We made a lot more changes, as this prompted us to (re)consider some popup logic and functionality.

Your commit was taken as the basis, with some more work on top of it.
As such, the PR will be closed (although it is "merged").

Best,
Matthias

tcld:
Thank you very much for the work and the update. And, since I did notice a few more cases where the popup behavior could have used some improvement, I'm looking forward to the next version!

PR #221, opened 2024-12-05
Fix doc/tutorial/auth.html 

ediw8311xht:
Include statement in tutorial is incorrect:
`#include <Wt/Auth/UserDatabase.h>`

Fixed to:
`#include <Wt/Auth/Dbo/UserDatabase.h>`


MatthiasVC:
That's indeed correct!

Sorry for the delay!

PR #222, opened 2024-12-11
Handle text-decoration:none option

TavakoliAli:
Hello

We could have created a zero(none) text-decoration, but it was un-handled and had no effect. (At least I think so :-) )

### DIFF
--- a/src/Wt/WCssDecorationStyle.C
+++ b/src/Wt/WCssDecorationStyle.C
@@ -370,14 +370,19 @@ void WCssDecorationStyle::updateDomElement(DomElement& element, bool all)
   if (textDecorationChanged_ ||  all) {
     std::string options;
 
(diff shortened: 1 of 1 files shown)


MatthiasVC:
Hello TavakoliAli,

apologies for the delay. I believe this is indeed incorrectly handled. However, more changes are required here.
With your extension `none` is always rendered, even in the default (so when not explicitly calling `setTextDecoration(Wt::None)`). That is too strong. The default should still be `textdecoration: ''`;, so that is inherits by default. Only when explicitly calling `setTextDecoration(Wt::TextDecoration::None)` should this be rendered out as such.

As more work is necessary I have created [WT-13699](https://redmine.emweb.be/issues/13699) to track this.
I will use your code as the basis, to leave credit.

However, I am closing the MR as this will pass internal review.

PR #225, opened 2025-01-18
fix: compatibility with boost-asio-1.87.0

miyanyan:
hi, asio removed some previously deprecated facilities, see https://www.boost.org/doc/libs/1_87_0/doc/html/boost_asio/history.html, I found this when I update vcpkg's boost to 1.87

base on https://github.com/emweb/wt/blob/master/doc/MinimumDependencyVersions.md, we need support boost>=1.66

so I made these changes, and all these functions can be found with boost-1.66:
* add io_service to `src\Wt\AsioWrapper\io_service.hpp`
* add rfc2818_verification to `src\Wt\AsioWrapper\ssl.hpp`
* remove tcp::resolver::query, use [ip::basic_resolver::resolve](https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/ip__basic_resolver/resolve.html)
* use [tcp::resolver::results_type::iterator](https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/ip__basic_resolver_results/iterator.html)
* use [asio::post](https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/post.html)
* use [asio::executor_work_guard<asio::io_service::executor_type>](https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/executor_work_guard.html)
* use [asio::io_context::restart](https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/io_context/restart.html)
* use [asio::steady_timer::expires_after](https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/basic_waitable_timer/expires_after.html)
* use [asio::ip::make_address](https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/ip__address/make_address.html)
* remove buffer_cast, use [data()](https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/buffer_cast.html)

### DIFF
--- a/src/Wt/AsioWrapper/asio.hpp
+++ b/src/Wt/AsioWrapper/asio.hpp
@@ -8,6 +8,7 @@
 #define WT_ASIO_ASIO_H_
 
 #include "Wt/WConfig.h"
+#include "io_service.hpp"
 
 #ifdef WT_ASIO_IS_BOOST_ASIO
 
--- a/src/Wt/AsioWrapper/io_service.hpp
+++ b/src/Wt/AsioWrapper/io_service.hpp
@@ -11,11 +11,29 @@
 
 #ifdef WT_ASIO_IS_BOOST_ASIO
 
+#include <boost/asio/version.hpp>
+#if BOOST_ASIO_VERSION >= 103300
+#include <boost/asio/io_context.hpp>
+namespace boost {
+  namespace asio {
+    using io_service = boost::asio::io_context;
+  }
+}
+#else
 #include <boost/asio/io_service.hpp>
+#endif
 
 #else // WT_ASIO_IS_STANDALONE_ASIO
 
+#include <asio/version.hpp>
+#if ASIO_VERSION >= 103300
+#include <asio/io_context.hpp>
+namespace asio {
+  using io_service = asio::io_context;
+}
+#else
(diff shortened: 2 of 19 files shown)


MatthiasVC:
Hello @miyanyan,

we had already been made aware of the issue. And as such, have a [ticket´](https://redmine.emweb.be/issues/13361), and already an open MR on our internal git. Sadly that means double work was performed.

Now, since the open MR on our end was already under review, and will go through our own pipeline and tests, I would prefer we keep that mainly as it is.

I do however very much like your approach of handling all version checking in the wrapping headers, and not anywhere in the actual code. That results in code that looks a lot cleaner. I have suggested to the one that implemented the change to take your work as the base, and put his work on top of it. So this PR will be closed, but the commit will be taken as the basis, so you will still be credited for your work.

I suspect you will be able to find the change on master today or tomorrow.

I hope you are happy with the in-between solution.

MatthiasVC:
Hey @miyanyan,

as promised, the change is now present on master, using your original commits.

All the best, and thank you for your work!

miyanyan:
thanks, waiting for your new release~

PR #231, opened 2025-06-23
std::ranges support for dbo::collections

Nuxar1:
Because the default constructor of std::collection<>::iterator is private, std::ranges::end() does not compile.
std::ranges::end requires std::semiregular. 
Making the default constructor public solves the issue.


RomainMard:
Thanks for the pr, it is Implemented in 4.13.0

PR #234, opened 2025-10-16
Add safeguards to avoid crashes

tcld:
When custom delegate derive from WItemDelegate and override createEditor, but neither of the two *editState methods, these can lead to hard crashes.
Like with updateModelIndex, now the methods check whether their cast pointers actually exist.


RomainMard:
Thanks for the pull request,

This will be added to Wt 4.13.0

PR #239, opened 2026-02-25
Add WTableView::visibleRowCount

m-seibt:
Add `int WTableView::visibleRowCount() const;`

Implements the most important part of https://redmine.emweb.be/issues/13787.


RomainMard:
Thanks for the pull request. However, we won't be using it as that function does not correctly return the amount of fully visible rows, but rather the maximum amount of rows that could be visible. That function will be available with the same name in 4.14.0

RomainMard:
This was added to version 4.14.0

PR #245, opened 2026-04-24
WT-14474: Align overloads of WDialog::positionAt and suppress centerDialog

m-seibt:
* `WDialog::positionAt(const WMouseEvent&)`: Show the dialog if hidden in order to
  - bring the position into effect
  - align the behavior with `WDialog::positionAt(const WWidget *, ...)`
* `WDialog::setHidden`: Suppress `wtObj.centerDialog()` if it has an x-offset (`setOffsets(..., Side::Left)`)
  which is the case if shown by both `WDialog::positionAt` overloads
  in order to avoid flicker with `WDialog::positionAt(const WMouseEvent&)`

### DIFF
--- a/src/Wt/WDialog.C
+++ b/src/Wt/WDialog.C
@@ -643,9 +643,9 @@ void WDialog::setHidden(bool hidden, const WAnimation& animation)
 
     DialogCover *c = cover();
     if (!hidden) {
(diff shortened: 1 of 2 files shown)


RomainMard:
Thanks for the pull request, we are going to add it to Wt soon

PR #246, opened 2026-06-16
WMenu: enforce segment boundaries in internal path matching

saddamr3e:
Problem

WMenu uses the internal match(path, component) helper function to determine how well the current internal path matches a menu item's path component.

The current implementation incorrectly treats prefix matches as valid even when they do not end on a path segment boundary.

For example:

If the current path is /contact-us and the menu contains a "contact" item, match("contact-us", "contact") returns 7, causing the "contact" menu item to be incorrectly selected even though /contact-us represents a different route.
Likewise, match("a/ab", "a/ac") returns 1 instead of -1, which contradicts the documented behavior of the helper function.
Solution
Update match() in src/Wt/WMenu.C to enforce path-segment boundary checks.
Return a positive match only when the component is an exact match or when the following character in the path is '/'.
Add the regression test WMenu_internal_path_matching_segment_boundary in test/widgets/WMenuTest.C to prevent future regressions.

### DIFF
--- a/src/Wt/WMenu.C
+++ b/src/Wt/WMenu.C
@@ -29,21 +29,21 @@ namespace {
    * ("a", "a") -> 1
    */
   int match(const std::string& path, const std::string& component) {
-    if (component.length() > path.length())
-      return -1;
-
-    int length = std::min(component.length(), path.length());
-
-    int current = -1;
-
-    for (int i = 0; i < length; ++i) {
-      if (component[i] != path[i]) {
-        return current;
-      } else if (component[i] == '/')
-        current = i;
+    if (component.empty()) {
+      return 0;
     }
 
-    return length;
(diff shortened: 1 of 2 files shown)


RomainMard:
Thanks for the pull request.

This change will be added to Wt soon.

Note that the test you added was incorrect and had to be fixed. This was because:

1. The first element added to the `WMenu` will be automaticaly selected, even if it does not match the internal path.
2. `WApplication::setInternalPath()` needs its second parameter to be set to `true` in order for the `WMenu` to react to the internal path change.

saddamr3e:
Thanks for the review. I have corrected the test as suggested and updated it accordingly.

saddamr3e:
any update?

RomainMard:
This has been added to Wt 4.14.0

PR #247, opened 2026-06-17
fix(WMenuItem): setContents() on the current item moves the selection to the previous item

saddamr3e:
While testing `WMenuItem::setContents()`, I noticed that updating the contents of the currently selected menu item moves the selection to the previous item.

`setContents()` reinserts the item via `removeItem()`/`insertItem()`. `removeItem()` moves the selection to the previous item, which is the intended behavior for removals, but the reinsert doesn't restore the selection, so a pure contents update ends up changing the current item.

This change leaves `removeItem()` untouched. `setContents()` now remembers whether the item was current and restores the selection after reinserting it. The restore updates `current_` before calling `select()` (the same pattern `removeItem()` uses internally), so no spurious `triggered()`/`itemSelected()` signals are emitted.

Includes a regression test.

### DIFF
--- a/src/Wt/WMenu.C
+++ b/src/Wt/WMenu.C
@@ -283,10 +283,13 @@ std::unique_ptr<WMenuItem> WMenu::removeItem(WMenuItem *item)
 
     item->setParentMenu(nullptr);
 
-    if (itemIndex <= current_ && current_ >= 0)
+    if (itemIndex == current_) {
+      setCurrent(-1);
+      select(-1, true);
+    } else if (itemIndex < current_ && current_ > 0) {
       --current_;
-
-    select(current_, true);
+      select(current_, true);
+    }
   }
 
   return result;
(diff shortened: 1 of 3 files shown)


saddamr3e:
any update?

RomainMard:
Hi saddamr3e,

Sorry for taking so long to answer.

We actually want the index to move to the previous value. So it is not possible to fix `WMenuItem::setContents()` that way.

saddamr3e:
Makes sense. I reworked it: removeItem() is back to its original behavior, and setContents() now remembers whether the item was current and restores the selection after reinserting it. The restore uses the same setCurrent-then-select pattern removeItem() uses internally, so it doesn't emit spurious triggered()/itemSelected(). Updated the test and PR description to match.

RomainMard:
This is only solving the issue when the item being changed is the current item. But this also happens when the `WMenuItem` changed is the in front of the selection. Additionally, the possibility of submenus also need to be taken into account. I ended up fixing it myself, since it required several changes.

Thanks for the work you did put in this.

saddamr3e:
Fair enough, mine only covered the case where the changed item is the current one, and I hadn't thought about items before the selection or submenus. Glad it got a proper fix, feel free to close this one.

PR #249, opened 2026-06-18
OSS-Fuzz: Add new fuzzer targets cgi parsing

arthurscchan:
This PR creates a new OSS-Fuzz fuzzer targets cgi functions.


RomainMard:
Thanks for the PR.

PR #252, opened 2026-06-22
fix(WDate): handle negative (BCE) years correctly

saddamr3e:
While testing WDate with negative (BCE) years, I noticed that dates before the common era do not behave correctly.

The date is packed internally in WDate::setYmd() by shifting the year value and storing the result in an unsigned field. For negative years, this leads to incorrect behavior, causing BCE dates to compare incorrectly and making year values inconsistent.

This change preserves negative years correctly when packing the date and stores the packed value in a signed field. It also updates the validity check so BCE dates are handled consistently.

I also added a regression test to cover BCE dates and ensure comparisons and year extraction continue to work correctly.

### DIFF
--- a/src/Wt/WDate.C
+++ b/src/Wt/WDate.C
@@ -124,10 +124,6 @@ void WDate::setDate(int year, int month, int day)
 
 void WDate::setYmd(int y, int m, int d)
 {
-  // Cast to unsigned before shifting to avoid UB on negative years (C++11/14/17).
-  // The two's-complement bit pattern is then stored in signed int ymd_ so that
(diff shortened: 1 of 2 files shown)


saddamr3e:
any update?

RomainMard:
Hi saddamr3e,

Thanks for the PR. Your fix will likely not be in the next release, but will be added in a future patch.

saddamr3e:
Sounds good, thanks for the update. No rush on my end.

RomainMard:
Added to version 4.14.1

